PR #23, opened 2022-09-21
Add priority to check tx based on gas fees

Kbhat1:
Based on https://github.com/cosmos/cosmos-sdk/pull/11985 and https://github.com/cosmos/cosmos-sdk/pull/12953

### DIFF
--- a/baseapp/baseapp.go
+++ b/baseapp/baseapp.go
@@ -662,7 +662,7 @@ func (app *BaseApp) runTx(ctx sdk.Context, mode runTxMode, txBytes []byte) (gInf
 
 	// only run the tx if there is block gas remaining
 	if mode == runTxModeDeliver && ctx.BlockGasMeter().IsOutOfGas() {
-		return gInfo, nil, nil, 0, sdkerrors.Wrap(sdkerrors.ErrOutOfGas, "no block gas left to run tx")
+		return gInfo, nil, nil, -1, sdkerrors.Wrap(sdkerrors.ErrOutOfGas, "no block gas left to run tx")
 	}
 
 	defer func() {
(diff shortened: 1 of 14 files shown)


baseapp/baseapp.go
```diff
@@ -662,7 +662,7 @@ func (app *BaseApp) runTx(ctx sdk.Context, mode runTxMode, txBytes []byte) (gInf
 
 	// only run the tx if there is block gas remaining
 	if mode == runTxModeDeliver && ctx.BlockGasMeter().IsOutOfGas() {
-		return gInfo, nil, nil, sdkerrors.Wrap(sdkerrors.ErrOutOfGas, "no block gas left to run tx")
+		return gInfo, nil, nil, 0, sdkerrors.Wrap(sdkerrors.ErrOutOfGas, "no block gas left to run tx")
```
codchen, line 665: do we want to set these error cases to -1 so that if some valid txs didn't specify priority and default to 0, at least they won't be ranked after the invalid ones?

x/auth/ante/fee.go
```diff
@@ -8,121 +8,96 @@ import (
...
+	if tfc == nil {
+		tfc = checkTxFeeWithValidatorMinGasPrices
+	}
+
 	return DeductFeeDecorator{
-		ak:             ak,
+		accountKeeper:  ak,
 		bankKeeper:     bk,
 		feegrantKeeper: fk,
+		txFeeChecker:   tfc,
```
codchen, line 35: if this field is required, we probably want to add it to `simapp` in this repo to 1 prevent it from breaking and 2 serve as an example

codchen (approved):

x/auth/ante/fee.go
Kbhat1: Added it in the simapp where we create a new AnteHandler in latest commit. However it is not required ```func NewDeductFeeDecorator(ak AccountKeeper, bk types.BankKeeper, fk FeegrantKeeper, tfc TxFeeChecker) DeductFeeDecorator {
	if tfc == nil {
		tfc = CheckTxFeeWithValidatorMinGasPrices
	}```

PR #25, opened 2022-09-23
Update StoreKey

BrandonWeng:
See comments


BrandonWeng:
Going to merge this to unblock dev in sei-chain and this is a pretty simple change

PR #27, opened 2022-09-26
Add Tx channels to context

BrandonWeng:
Adds transaction channels to context 

Had to shuffle around to avoid circular dependency but essentially moved access ops and mapDependency objects to types/, while genesis info remains in x/accessmodule/types

This PR is used in https://github.com/sei-protocol/sei-chain/pull/261


types/context.go
```diff
@@ -39,6 +39,8 @@ type Context struct {
 	consParams    *tmproto.ConsensusParams
 	eventManager  *EventManager
 	priority      int64 // The tx priority, only relevant in CheckTx
+	txCompletionChannels [][]chan interface{}
```
udpatil, line 42: how will we be using these to know when to block / signal on resource need / release

BrandonWeng: Updated it to be a map of aclops -> []channels

types/context.go
```diff
@@ -39,8 +40,18 @@ type Context struct {
 	consParams    *tmproto.ConsensusParams
 	eventManager  *EventManager
 	priority      int64 // The tx priority, only relevant in CheckTx
+
+	txBlockingChannels		MessageAccessOpsChannelMapping
+	txCompletionChannels	MessageAccessOpsChannelMapping
+	messageIndex int	// Used to track current message being processed
```
udpatil, line 46: why are we tracking message index instead of txindex? in the other places for dependency mapping, we're indexing by the tx index, this distinction is relevant since each tx can have multiple messages as part of it.

types/context.go
```diff
@@ -39,8 +40,18 @@ type Context struct {
...
 }
 
+// Alias for Map of MessageIndex -> AccessOperation -> Channel
+type MessageAccessOpsChannelMapping = map[int]AccessOpsChannelMapping
```
udpatil, line 50: this should map by tx index

types/tx_msg.go
```diff
@@ -78,4 +78,4 @@ type TxEncoder func(tx Tx) ([]byte, error)
 // MsgTypeURL returns the TypeURL of a `sdk.Msg`.
 func MsgTypeURL(msg Msg) string {
 	return "/" + proto.MessageName(msg)
-}
+}
```
udpatil, line 81: nit: newline

philipsu522 (approved):

udpatil (approved):

PR #29, opened 2022-09-28
Add msg handler wrapper for concurrent processing 

BrandonWeng:
Creating a wrapper that gets called before all message handlers

If the completion signals are empty then it should not block anything as the slices of channels will be empty.

### DIFF
--- a/types/accesscontrol/access_operation_map.go
+++ b/types/accesscontrol/access_operation_map.go
@@ -6,7 +6,6 @@ type MessageAccessOpsChannelMapping = map[int]AccessOpsChannelMapping
 // Alias for Map of AccessOperation -> Channel
 type AccessOpsChannelMapping = map[*AccessOperation][]chan interface{}
 
-// Waits for all signals to complete before proceeding
 func WaitForSignals(accessOpsToChannelsMap AccessOpsChannelMapping) {
 	for _, channels := range accessOpsToChannelsMap {
 		for _, channel := range channels {
--- a/types/accesscontrol/access_operation_map.go
+++ b/types/accesscontrol/access_operation_map.go
@@ -1,0 +1,24 @@
+package accesscontrol
+
+// Alias for Map of MessageIndex -> AccessOperation -> Channel
+type MessageAccessOpsChannelMapping = map[int]AccessOpsChannelMapping
+
+// Alias for Map of AccessOperation -> Channel
+type AccessOpsChannelMapping = map[*AccessOperation][]chan interface{}
+
+// Waits for all signals to complete before proceeding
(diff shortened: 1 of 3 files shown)


types/accesscontrol/access_operation_map.go
```diff
@@ -0,0 +1,23 @@
+package accesscontrol
+
+// Alias for Map of MessageIndex -> AccessOperation -> Channel
+type MessageAccessOpsChannelMapping = map[int]AccessOpsChannelMapping
+
+// Alias for Map of AccessOperation -> Channel
+type AccessOpsChannelMapping = map[*AccessOperation][]chan interface{}
```
udpatil, line 7: Why is this definition a map on struct pointer? I thought we were keying on the struct itself previously?

types/accesscontrol/access_operation_map.go
```diff
@@ -0,0 +1,23 @@
...
+
+func WaitForSignals(accessOpsToChannelsMap AccessOpsChannelMapping) {
+	for _, channels := range accessOpsToChannelsMap {
+		for _, channel := range channels {
+			<-channel
+		}
+	}
+}
+
+func SendSignals(accessOpsToChannelsMap AccessOpsChannelMapping) {
```
udpatil, line 17: Let's call this something like SendAllSignals since we'll likely have one later that also takes a specific access op to signal for

types/accesscontrol/access_operation_map.go
```diff
@@ -0,0 +1,23 @@
...
+	for accessOp, channels := range accessOpsToChannelsMap {
+		for _, channel := range channels {
+			channel <- accessOp
```
udpatil, line 20: I believe we can just write an empty struct to the chann for more efficiency struct{}{}

types/accesscontrol/access_operation_map.go
BrandonWeng: sg, updated th other one to also be called waitforallsignals

types/accesscontrol/access_operation_map.go
BrandonWeng: ah good point, updated

types/accesscontrol/access_operation_map.go
BrandonWeng: Sorry, forgot to include the context.go change, but it was using pointer as the key. It shouldn't though as we may need to construct it later on if we want more granular access ops

https://github.com/sei-protocol/sei-cosmos/pull/29/files#diff-10fabc84ecc74a1e1f8383ab56fb83d48a5528658a27ee5373d05a13d294d438L54

philipsu522 (approved):

udpatil (approved):

PR #30, opened 2022-09-29
[accesscontrol] Add resource hierarchy helper

udpatil:
This adds in a hardcoded resource hierarchy along with the helper to build a list of resource dependencies based on an initial resource.

### DIFF
--- a/types/accesscontrol/resource.go
+++ b/types/accesscontrol/resource.go
@@ -1,0 +1,41 @@
+package accesscontrol
+
+type TreeNode struct {
+	Parent   ResourceType
+	Children []ResourceType
+}
+
+// This returns a slice of all resource types that are dependent to a specific resource type
+// Travel up the parents to add them all, and add all children as well.
+// TODO: alternatively, hardcode all dependencies (parent and children) for a specific resource type, so we don't need to do a traversal when building dag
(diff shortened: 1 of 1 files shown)


types/accesscontrol/resource.go
```diff
@@ -0,0 +1,41 @@
...
+	Children []ResourceType
+}
+
+// This returns a slice of all resource types that are dependent to a specific resource type
+// Travel up the parents to add them all, and add all children as well.
+// TODO: alternatively, hardcode all dependencies (parent and children) for a specific resource type, so we don't need to do a traversal when building dag
+func (r ResourceType) GetResourceDependencies() []ResourceType {
+	// resource is its own dependency
+	resources := []ResourceType{r}
+	resourceTree := map[ResourceType]TreeNode{
```
BrandonWeng, line 14: Does it make sense to make this a field like resource dependency mapping? This way we can change it through a proposal too

udpatil: I don't necessarily think it makes sense because to change it to add new fields and stuff, we'd need to make a code change to sei-cosmos anyways, so the resource types arent exactly extensible on chain via proposals

BrandonWeng (approved):

PR #31, opened 2022-10-03
Add default value if acl keymap doesnt exist

BrandonWeng:
Need to add  default resource access list if dep the msg key doesn't exist or it's not defined, this way it still just uses synchronous processing 

Ran load testing script with https://github.com/sei-protocol/sei-chain/compare/2.0.0beta...bweng-2.0.0beta-testing and verified that the DAG is built properly

### DIFF
--- a/x/accesscontrol/keeper/keeper.go
+++ b/x/accesscontrol/keeper/keeper.go
@@ -43,9 +43,14 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
 
 func (k Keeper) GetResourceDependencyMapping(ctx sdk.Context, messageKey string) acltypes.MessageDependencyMapping {
 	store := ctx.KVStore(k.storeKey)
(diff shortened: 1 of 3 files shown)


x/accesscontrol/keeper/keeper.go
```diff
@@ -43,9 +43,14 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
 
 func (k Keeper) GetResourceDependencyMapping(ctx sdk.Context, messageKey string) acltypes.MessageDependencyMapping {
 	store := ctx.KVStore(k.storeKey)
-	b := store.Get(types.GetResourceDependencyKey(messageKey))
+	depMapping := store.Get(types.GetResourceDependencyKey(messageKey))
+	if depMapping == nil {
```
BrandonWeng, line 47: without the default the object returns an empty list

philipsu522 (approved):

udpatil (approved):

PR #32, opened 2022-10-03
Add Query command to acl module

BrandonWeng:
Adding support for querying params and dependencies for acl module
image

Will follow up with another PR for governance proposals


proto/cosmos/accesscontrol_x/genesis.proto
```diff
@@ -1,5 +1,5 @@
 syntax = "proto3";
-package cosmos.accesscontrol_genesis.v1beta1;
+package cosmos.accesscontrol_x.v1beta1;
```
BrandonWeng, line 2: Renaming the package since we need the query, gov and genesis protos to be in the /x/ module to avoid circular imports

philipsu522 (approved):

PR #35, opened 2022-10-05
[acl] Add dependency switch to support dynamic deps

udpatil:
This introduces the `GetMessageDependencies` function that returns the access operations for the message. It either returns the static access ops from the keeper store, or if we've implemented a specific case for that message type, it will generate the dynamic message typing if applicable.

Need to refactor sei-chain with this function and verify the default behavior is as expected.


BrandonWeng (approved):

x/accesscontrol/keeper/keeper.go
```diff
@@ -130,11 +148,29 @@ func MeasureBuildDagDuration(start time.Time, method string) {
...
+			validateErr := types.ValidateAccessOps(dependencies)
+			if validateErr == nil {
+				return dependencies
+			} else {
+				ctx.Logger().Error(validateErr.Error())
+			}
+		} else {
+			ctx.Logger().Error("Error generating message dependencies: ", err)
+		}
+	}
```
philipsu522, line 167: would it be possible to add unit tests for this?

philipsu522 (approved):

PR #38, opened 2022-10-07
Move Parallel TX Signalling to Sei-Chain DeliverTx call 

BrandonWeng:
Same as this PR: https://github.com/sei-protocol/sei-chain/pull/291

Previously we were signaling and blocking for each handler in the message - in theory, this is fine as we are supposed to be processed synchronously based on the access operation, however, due to us not having enough granularity, blocking on messages is not sufficient as the ordering for when each message is unblocked (between different TXs) could be different. This results in different gas fees or different bank AVL tree shapes as the node insertion order matters. 

Therefore, as we do not have more granular support for per-message concurrency, we should be okay with just blocking on the entire transaction and ensuring that no r/w can occur unless the entire TX is ready to be processed. This is operating on the assumption that the majority of transactions will not have overlapping resources. 

Note that the Gas used is based on the size of the data returned, depending on the ordering of the message process it's possible that some may result in the same bank data being returned with less digits, and therefore costing different gas. 

Ran the 20 iterations of load test with 5 rounds each and saw that the chain didn't get stuck and from previous outputs of the delivered, the messages are being processed sequentially as expected based on the logs. This was in a 5 node cluster

With these PRs merged, we should be able to then define more granular access operations for the message types we want to support e.g Bank Send should have access operations for read to sender, writing to both sender and receiver, and the contractAddr 
https://github.com/sei-protocol/sei-cosmos/pull/35
https://github.com/sei-protocol/sei-cosmos/pull/37
https://github.com/sei-protocol/sei-chain/pull/287

### DIFF
--- a/types/accesscontrol/access_operation_map.go
+++ b/types/accesscontrol/access_operation_map.go
@@ -6,18 +6,22 @@ type MessageAccessOpsChannelMapping = map[int]AccessOpsChannelMapping
 // Alias for Map of AccessOperation -> Channel
 type AccessOpsChannelMapping = map[AccessOperation][]chan interface{}
 
-func WaitForAllSignals(accessOpsToChannelsMap AccessOpsChannelMapping) {
-	for _, channels := range accessOpsToChannelsMap {
-		for _, channel := range channels {
-			<-channel
+func SendAllSignals(txIndex int, messageIndexToAccessOpsChannelMapping MessageAccessOpsChannelMapping) {
+	for _, accessOpsToChannelsMap  := range messageIndexToAccessOpsChannelMapping {
+		for _, channels := range accessOpsToChannelsMap {
+			for _, channel := range channels {
+				channel <- struct{}{}
+			}
 		}
 	}
 }
 
-func SendAllSignals(accessOpsToChannelsMap AccessOpsChannelMapping) {
-	for _, channels := range accessOpsToChannelsMap {
-		for _, channel := range channels {
-			channel <- struct{}{}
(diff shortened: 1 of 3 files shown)


types/accesscontrol/access_operation_map.go
```diff
@@ -6,18 +6,22 @@ type MessageAccessOpsChannelMapping = map[int]AccessOpsChannelMapping
...
+			}
 		}
 	}
 }
 
-func SendAllSignals(accessOpsToChannelsMap AccessOpsChannelMapping) {
-	for _, channels := range accessOpsToChannelsMap {
-		for _, channel := range channels {
-			channel <- struct{}{}
+func WaitForAllSignals(txIndex int, messageIndexToAccessOpsChannelMapping MessageAccessOpsChannelMapping) {
```
udpatil, line 19: what is txIndex used for?

types/accesscontrol/access_operation_map.go
```diff
@@ -6,18 +6,22 @@ type MessageAccessOpsChannelMapping = map[int]AccessOpsChannelMapping
 // Alias for Map of AccessOperation -> Channel
 type AccessOpsChannelMapping = map[AccessOperation][]chan interface{}
 
-func WaitForAllSignals(accessOpsToChannelsMap AccessOpsChannelMapping) {
-	for _, channels := range accessOpsToChannelsMap {
-		for _, channel := range channels {
-			<-channel
+func SendAllSignals(txIndex int, messageIndexToAccessOpsChannelMapping MessageAccessOpsChannelMapping) {
```
udpatil, line 9: what is txIndex used for?

PR #40, opened 2022-10-07
Move blocking to beginning of runTx

BrandonWeng:
There are resource accesses in the ante handler, we need to block before the access occurs, or else we will run into issues there where there's a race condition between the different accesses

Ran 20 iterations of the the basic msg load test with 5 validators and didn't experience any issues


philipsu522 (approved):
Nice!!!

philipsu522:
is there any way to add a unit test for this? i.e. ensure that lock is held before any gas deducted / any toher function calls? not a blocker though - we can revist

PR #41, opened 2022-10-07
Ante deps

udpatil:
This introduces an ante dependency decorator and an interface by which we can define the dependencies for ante handlers.

Needs testing with the related sei-chain change


philipsu522 (approved):
lgtm, but might be good to wait for a second review

BrandonWeng (approved):

PR #54, opened 2022-10-13
Bweng acl order fix

BrandonWeng:

### DIFF
--- a/x/accesscontrol/keeper/keeper.go
+++ b/x/accesscontrol/keeper/keeper.go
@@ -227,7 +227,7 @@ func (k Keeper) GetMessageDependencies(ctx sdk.Context, msg sdk.Msg) []acltypes.
 	// Default behavior is to get the static dependency mapping for the message
 	messageKey := types.GenerateMessageKey(msg)
 	dependencyMapping := k.GetResourceDependencyMapping(ctx, messageKey)
-	if dependencyGenerator, ok := k.MessageDependencyGeneratorMapper[types.GenerateMessageKey(msg)]; ok {
+	if dependencyGenerator, ok := k.MessageDependencyGeneratorMapper[types.GenerateMessageKey(msg)]; dependencyMapping.DynamicEnabled && ok {
 		// if we have a dependency generator AND dynamic is enabled, use it
 		if dependencies, err := dependencyGenerator(k, ctx, msg); err == nil {
 			// validate the access ops before using them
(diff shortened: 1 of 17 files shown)


x/auth/ante/basic.go
```diff
@@ -79,9 +80,29 @@ type ConsumeTxSizeGasDecorator struct {
 	ak AccountKeeper
 }
 
-func NewConsumeGasForTxSizeDecorator(ak AccountKeeper) ConsumeTxSizeGasDecorator {
-	return ConsumeTxSizeGasDecorator{
-		ak: ak,
+type ConsumeTxSizeGasDepDecorator struct {}
+
+func (d ConsumeTxSizeGasDepDecorator) AnteDeps(txDeps []sdkacltypes.AccessOperation, tx sdk.Tx, next sdk.AnteDepGenerator) (newTxDeps []sdkacltypes.AccessOperation, err error) {
```
udpatil, line 85: We can just implement this for `ConsumeTxSizeGasDecorator` and then we don't have to specifically wrap them together.

x/accesscontrol/keeper/keeper.go
```diff
@@ -227,7 +227,7 @@ func (k Keeper) GetMessageDependencies(ctx sdk.Context, msg sdk.Msg) []acltypes.
 	// Default behavior is to get the static dependency mapping for the message
 	messageKey := types.GenerateMessageKey(msg)
 	dependencyMapping := k.GetResourceDependencyMapping(ctx, messageKey)
-	if dependencyGenerator, ok := k.MessageDependencyGeneratorMapper[types.GenerateMessageKey(msg)]; dependencyMapping.DynamicEnabled && ok {
+	if dependencyGenerator, ok := k.MessageDependencyGeneratorMapper[types.GenerateMessageKey(msg)]; ok {
```
udpatil, line 230: why do we want to remove `dynamic enabled`? or just for testing?

x/auth/ante/basic.go
```diff
@@ -79,9 +80,29 @@ type ConsumeTxSizeGasDecorator struct {
...
+				ResourceType:       sdkacltypes.ResourceType_KV,
+				IdentifierTemplate:  signer.String(),
+			},
+		)
+
+	}
+	return next(append(txDeps, deps...), tx)
+}
+
+func NewConsumeGasForTxSizeDecorator(ak AccountKeeper) sdk.WrappedAnteDecorator {
```
udpatil, line 102: here we can instead just return `ConsumeTxSizeGasDecorator` and implement AnteDeps for that struct

x/accesscontrol/keeper/keeper.go
BrandonWeng: oh ye, this is just for testing - instead of enabling it everytime through gov proposals

PR #58, opened 2022-10-17
Lazy Deposit All Module Accounts During EndBlock

BrandonWeng:
Relevant PR:
https://github.com/sei-protocol/sei-chain/pull/313

We are moving the module deposits to the end of the block and transferring them all at once. This is to prevent the gas fee deposit from being a single source of a bottleneck when it comes to concurrency. 

Ran 3 iterations of load tests with 5 rounds, and 400 orders per block 
image

image


x/accesscontrol/types/graph.go
```diff
@@ -130,6 +130,12 @@ func (dag *Dag) AddEdge(fromIndex DagNodeID, toIndex DagNodeID) *DagEdge {
 	if _, ok := dag.NodeMap[toIndex]; !ok {
 		return nil
 	}
+
+	// Processing in a given TX is sequential so no need for dependency
```
udpatil, line 134: why are we adding this check here? AddEdge is meant to be a helper that solely adds an edge, whereas helpers such as `AddNodeBuildDependency` will do the business logic of ensuring there arent edges within the same tx? I guess its fine to have this as an invariant for edge creation

x/bank/keeper/keeper.go
```diff
@@ -355,6 +367,56 @@ func (k BaseKeeper) SendCoinsFromAccountToModule(
 	return k.SendCoins(ctx, senderAddr, recipientAcc.GetAddress(), amt)
 }
 
+
+// LazyDepositFromAccountToModule transfers coins from an AccAddress to a ModuleAccount.
+// It deducts the balance from an accAddress and stores the balance in a mapping for ModuleAccounts.
+// In the EndBlocker, it will then perform one deposit for each module account.
+// It will panic if the module account does not exist.
+func (k BaseKeeper) LazySendCoinsFromAccountToModule(
```
udpatil, line 375: Instead of calling this LazySend, maybe we can call it something like `DeferredSend` to make more contextual sense that it performs the send at the end block?

x/bank/keeper/keeper.go
```diff
@@ -355,6 +367,56 @@ func (k BaseKeeper) SendCoinsFromAccountToModule(
...
+
+	newAmount := amount
+	if v, ok := k.moduleAccountDepositMapping[recipientModule]; ok {
+		newAmount = v.Add(amount...)
+	}
+	k.moduleAccountDepositMapping[recipientModule] = newAmount
+}
+
+// Iterates on all the lazy deposits and deposit them into the store
+func (k BaseKeeper) WriteLazyDepositsToModuleAccounts(ctx sdk.Context) []abci.Event {
```
udpatil, line 400: don't we also need to call this in the bank `endblock` logic?

BrandonWeng: I was thinking of calling it in the sei-chain app - the baseapp in sei-cosmos doesn't have access to the bank keeper

baseapp/block_gas_test.go
```diff
@@ -125,16 +125,17 @@ func TestBaseApp_BlockGas(t *testing.T) {
 				require.Equal(t, uint32(0), rsp.TxResults[0].Code)
 				require.Equal(t, []byte("ok"), okValue)
 			}
-			// check block gas is always consumed
-			baseGas := uint64(59142) // baseGas is the gas consumed before tx msg
+			// check block gas is always consumed - this value may change if we update the logic for
+			// how gas is consumed
+			baseGas := uint64(52585) // baseGas is the gas consumed before tx msg
```
BrandonWeng, line 130: We no longer charge everytime so the base gas fee goes down too

codchen (approved):

store/types/gas.go
```diff
@@ -52,26 +53,37 @@ type GasMeter interface {
 type basicGasMeter struct {
 	limit    Gas
 	consumed Gas
+	lock *sync.Mutex
```
BrandonWeng, line 56: BlockGas can/will be modified concurrently

philipsu522: would it make sense to make Gas an atomic by using
`atomic.AddUint64`  and `atomic. AddUint64`? This way if we add / remove any functions that touch gas, we know that the locking is already handled at a lower layer. O/w we need to be careful of ensuring every function that accesses any `Gas` variable handles the mutex \cc @codchen . Not a blocker, though

x/bank/keeper/keeper.go
```diff
@@ -355,6 +369,69 @@ func (k BaseKeeper) SendCoinsFromAccountToModule(
 	return k.SendCoins(ctx, senderAddr, recipientAcc.GetAddress(), amt)
 }
 
+
+// DeferredSendCoinsFromAccountToModule transfers coins from an AccAddress to a ModuleAccount.
+// It deducts the balance from an accAddress and stores the balance in a mapping for ModuleAccounts.
+// In the EndBlocker, it will then perform one deposit for each module account.
+// It will panic if the module account does not exist.
```
philipsu522, line 376: does the entire block get rolled back? IIRC it does, but wanted to double check

philipsu522 (approved):

x/bank/keeper/keeper.go
BrandonWeng: It would get stuck, but the block will not be committed - in this case we would be able upload a new binary to fix it.

PR #62, opened 2022-10-21
Use buffered channel for completion signal

codchen:
There are two types of Go channels: unbuffered and buffered. Buffered channel is created with an additional capacity argument to `make`. The difference between buffered and unbuffered channel is that if another goroutine starts reading the channel before the writing goroutine writes to it, the write will be blocked in the unbuffered channel's case until the reading goroutine gives up reading, whereas a buffered channel will allow the write.

Also backfilled some resource mappings.

Tested with wasm execute messages. Deadlock with unbuffered channel but function properly with buffered channel

### DIFF
--- a/types/accesscontrol/access_operation_map.go
+++ b/types/accesscontrol/access_operation_map.go
@@ -7,7 +7,7 @@ type MessageAccessOpsChannelMapping = map[int]AccessOpsChannelMapping
 type AccessOpsChannelMapping = map[AccessOperation][]chan interface{}
 
 func WaitForAllSignalsForTx(messageIndexToAccessOpsChannelMapping MessageAccessOpsChannelMapping) {
-	for _, accessOpsToChannelsMap  := range messageIndexToAccessOpsChannelMapping {
+	for _, accessOpsToChannelsMap := range messageIndexToAccessOpsChannelMapping {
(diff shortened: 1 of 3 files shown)


types/accesscontrol/access_operation_map.go
```diff
@@ -7,22 +7,23 @@ type MessageAccessOpsChannelMapping = map[int]AccessOpsChannelMapping
 type AccessOpsChannelMapping = map[AccessOperation][]chan interface{}
 
 func WaitForAllSignalsForTx(messageIndexToAccessOpsChannelMapping MessageAccessOpsChannelMapping) {
-	for _, accessOpsToChannelsMap  := range messageIndexToAccessOpsChannelMapping {
+	for _, accessOpsToChannelsMap := range messageIndexToAccessOpsChannelMapping {
 		for _, channels := range accessOpsToChannelsMap {
 			for _, channel := range channels {
 				<-channel
+				// reenqueue so that other reads can also be unblocked
+				channel <- struct{}{}
```
udpatil, line 15: why do we need to reenqueue? each dag edge has its own channel so we never need to write to a channel twice.

udpatil (approved):

BrandonWeng (approved):
thanks for the change and the detailed description!

PR #63, opened 2022-10-21
Bweng mint msg

BrandonWeng:

### DIFF
--- a/types/deferred_bank_operations.go
+++ b/types/deferred_bank_operations.go
@@ -1,0 +1,88 @@
+package types
+
+import (
+	"log"
+	"sort"
+	"sync"
+)
+
+type DeferredBankOperationMapping struct {
+	deferredOperations map[string]Coins
+	mappingLock		  *sync.Mutex
+}
+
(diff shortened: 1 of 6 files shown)


types/deferred_bank_operations.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func NewDeferredBankOperationMap() *DeferredBankOperationMapping {
+	return &DeferredBankOperationMapping{
+		deferredOperations: make(map[string]Coins),
+		mappingLock: &sync.Mutex{},
+	}
+}
+
+func (m *DeferredBankOperationMapping) get(moduleAccount string) (Coins, bool) {
```
philipsu522, line 20: these functions don't need locking?

BrandonWeng: I made these private functions as they should only be called internally. let me add comments

BrandonWeng: Actually, i just removed this. It's not really needed and adds more confusion

philipsu522 (approved):

PR #67, opened 2022-10-22
Set default access ops for wasm contracts

codchen:
Right now if a wasm contract doesn't have access ops specified, it will cause all wasm messages to have parallel processing disabled because it will return an error. To prevent this we will return a default mapping for wasm messages instead.

tested with multiple contracts (one specified dependency and one doesn't)


udpatil (approved):

BrandonWeng (approved):
do we still need this?

PR #70, opened 2022-10-25
Validate Concurrent Messages + Update BankSend 

BrandonWeng:


store/cachekv/store.go
```diff
@@ -70,6 +78,7 @@ func (store *Store) Get(key []byte) (value []byte) {
 	} else {
 		value = cacheValue.value
 	}
+	store.eventManager.EmitResourceAccessReadEvent("get", key, value)
```
BrandonWeng, line 81: For deliver TX we technically only need to add it to the CacheKvStore when validating KVStore objects since we branch to two different CacheKvStores for anteHandler and runMsgs

types/events.go
```diff
@@ -73,6 +88,30 @@ func (em *EventManager) EmitTypedEvents(tevs ...proto.Message) error {
...
 }
 
+func (em *EventManager) EmitResourceAccessReadEvent(operation string, key, value []byte) {
+	em.EmitEvent(
+		NewEvent(
+			EventTypeResourceAccess,
+			NewAttribute(AttributeKeyAccessType, AttributeKeyAccessTypeRead),
+			NewAttribute(AttributeKeyResourceKey, conv.UnsafeBytesToStr(key)),
+			NewAttribute(AttributeKeyResourceValue, conv.UnsafeBytesToStr(value)),
+			NewAttribute(AttributeKeyOperation, operation),
```
BrandonWeng, line 98: Addtional info useful for debugging

types/events.go
```diff
@@ -73,6 +87,30 @@ func (em *EventManager) EmitTypedEvents(tevs ...proto.Message) error {
...
 }
 
+func (em *EventManager) EmitResourceAccessReadEvent(operation string, key, value []byte) {
+	em.EmitEvent(
+		NewEvent(
+			EventTypeResourceAccess,
+			NewAttribute(AttributeKeyAccessType, AttributeKeyAccessTypeRead),
+			NewAttribute(AttributeKeyResourceKey, string(key)),
+			NewAttribute(AttributeKeyResourceValue, string(value)),
+			NewAttribute(AttributeKeyOperation, operation),
```
BrandonWeng, line 97: Addtional info useful for debugging

codchen:
@BrandonWeng  one broader comment I have that i forgot to mention in the doc is if all we need the events for is to perform validation at the end of ProcessBlock, then we might want to use some other media than ABCI events, since ABCI events will be indexed to persistent storage on tendermint-level, and we have seen indexing too many events hurting chain performance in the past.

PR #75, opened 2022-10-27
Add multiplier gas meter

codchen:
Add a new gas meter type that allows passing in a multiplier. To ensure computation stability we represent multiplier in two ints (nominator and denominator). This PR also extended SetUpContextDecorator so that applications can have custom logic to decide whether to use multiplier gas meter and what multiplier to use.

unit test

### DIFF
--- a/store/types/gas.go
+++ b/store/types/gas.go
@@ -53,7 +53,7 @@ type GasMeter interface {
 type basicGasMeter struct {
 	limit    Gas
 	consumed Gas
-	lock *sync.Mutex
+	lock     *sync.Mutex
 }
 
 // NewGasMeter returns a reference to a new basicGasMeter.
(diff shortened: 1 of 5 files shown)


BrandonWeng (approved):

store/types/gas.go
```diff
@@ -149,16 +149,46 @@ func (g *basicGasMeter) String() string {
...
+}
+
+func NewMultiplierGasMeter(limit Gas, multiplierNominator uint64, multiplierDenominator uint64) GasMeter {
+	return &multiplierGasMeter{
+		basicGasMeter: basicGasMeter{
+			limit:    limit,
+			consumed: 0,
+			lock:     &sync.Mutex{},
+		},
+		multiplierNominator:   multiplierNominator,
```
udpatil, line 165: nit, this should be `Numerator`

codchen: ah good call

PR #76, opened 2022-10-27
Validate Concurrent Messages + Update BankSend 

BrandonWeng:
* CacheKv Stores emit an access operation event used to track accesses to resources during runTx, we only need to add it to cacheKv because we branch before runMsgs and ante handlers
* Adds validation logic 

Will also need to add validation for ante handlers 

Testing part of this PR:
https://github.com/sei-protocol/sei-chain/pull/345


types/accesscontrol/validation_test.go
```diff
@@ -0,0 +1,197 @@
...
+									{Key: "access_type", Value: "write"},
+								},
+							},
+						},
+					},
+			want:   map[Comparator]bool{
+				{AccessType: AccessType_WRITE, Identifier: "a/b/c/d/e"}: true,
+			},
+		},
+		{
```
BrandonWeng, line 161: Note that we're not checking for extra access ops defined, I think it's harder to check here since the path will not always result in the same set of access ops (e.g if account already exists vs if bank sends needs to create one)

I think this is fine for now as that's less f a priority and may just result in less concurrency but it's safer.

types/accesscontrol/validation.go
```diff
@@ -0,0 +1,118 @@
...
+			AccessType: accessType,
+			Identifier: identifier,
+		})
+	}
+	return comparators
+}
+
+// ValidateAccessOperations compares a list of events and a predefined list of access operations and determines if all the
+// events that occurred are represented in the accessOperations
+func ValidateAccessOperations(accessOps []AccessOperation, events []abci.Event) map[Comparator]bool {
```
codchen, line 88: i might have missed it but where do we compare resource type? it doesn't seem like we are consolidating all those different KV resource types

baseapp/baseapp.go
```diff
@@ -829,10 +829,12 @@ func (app *BaseApp) runMsgs(ctx sdk.Context, msgs []sdk.Msg, mode runTxMode) (*s
 			err          error
 		)
 
+		msgCtx, msgMsCache := app.cacheTxContext(ctx, []byte{})
```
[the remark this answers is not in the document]
BrandonWeng: The messages should not be committed in general as it's not being returned in the Tx result, but this branch allows us to make sure that the events in the CacheKv store only reflects the access ops for this message, so we can compare it per message, since that's how access ops are defined

store/cachekv/memiterator.go
```diff
@@ -51,6 +54,10 @@ func (mi *memIterator) Value() []byte {
 	if _, ok := mi.deleted[string(key)]; ok && !reCallingOnOldLastKey {
 		return nil
 	}
+
+	value := mi.Iterator.Value()
+	mi.eventManager.EmitResourceAccessReadEvent("iterator", key, value)
```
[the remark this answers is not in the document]
BrandonWeng: Not for the validation but I found it to be really useful for debugging missing Access Ops as most of the time the keys are not in a readable format. 

We're also not committing the events returned from this store so I thought it would be fine to leave it here

types/accesscontrol/validation.go
BrandonWeng: Addressed in the follow up commits, passing a store key, and defining a mappign of the storekey to the resource types and prefixes to do the matching

types/accesscontrol/comparator.go
```diff
@@ -0,0 +1,110 @@
...
+			AccessType: accessType,
+			Identifier: identifier,
+			StoreKey: storeKey,
+
+		})
+	}
+	return comparators
+}
+
+func (c *Comparator) DependencyMatch(accessOp AccessOperation, prefix []byte) bool {
```
BrandonWeng, line 68: in the process of adding more unit tests for this method

codchen (approved):

PR #77, opened 2022-10-28
Fixing Banksend unit tests 

BrandonWeng:
With the deferred logic, we need to also check the pending deposits when subtracting coins from the module accounts, this was added to some of the methods, but some other areas were missed 

image

### DIFF
--- a/x/bank/keeper/keeper.go
+++ b/x/bank/keeper/keeper.go
@@ -357,15 +357,19 @@ func (k BaseKeeper) DeferredSendCoinsFromModuleToAccount(
 		return sdkerrors.Wrapf(sdkerrors.ErrUnauthorized, "%s is not allowed to receive funds", recipientAddr)
 	}
 
-	// Branch Context for validation and fail if the module doesn't have enough coins
-	// but don't write this to the underlying store
-	validationContext, _ := ctx.CacheContext()
-	err := k.subUnlockedCoins(validationContext, moduleAddr, amount)
-	if err != nil {
-		return err
+	// Subtract from any pending sends fist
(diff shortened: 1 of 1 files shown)


x/bank/keeper/keeper.go
```diff
@@ -619,6 +623,14 @@ func (k BaseKeeper) destroyCoins(ctx sdk.Context, moduleName string, amounts sdk
 // It will panic if the module account does not exist or is unauthorized.
 func (k BaseKeeper) BurnCoins(ctx sdk.Context, moduleName string, amounts sdk.Coins) error {
 	subFn := func(ctx sdk.Context, moduleName string, amounts sdk.Coins) error {
+
+		// Try subtract from the in mem var first, prevents the condition where
+		// the module may have a pending deposit that would be enough to pay for this send
+		ok := ctx.ContextMemCache().SafeSubDeferredSends(moduleName, amounts)
```
udpatil, line 629: is it safe to add this in the normal BurnCoins and not the DeferredBurn? Doesn't that make the deferred mem resources a potential race condition?

BrandonWeng: This is in the case where we call deferredMint and then call the regular burn, in this case if we don't check the deferred sends then it will error out with insufficient balance. 

The deferred in mem variables are all blocked by mutex, I think in this case we only care about the absolute value of the in-mem variables at the end of FinalizeBlocker, as only a single value is written to the store

udpatil: gotcha, makes sense

udpatil (approved):

PR #81, opened 2022-11-01
Update ante handler dependencies

BrandonWeng:
Updates ante handler dependencies 

https://github.com/sei-protocol/sei-chain/pull/354

### DIFF
--- a/types/handler.go
+++ b/types/handler.go
@@ -153,5 +153,6 @@ func (t Terminator) AnteHandle(ctx Context, _ Tx, _ bool, _ AnteHandler) (Contex
 
 // Simply return provided txDeps and nil error
 func (t Terminator) AnteDeps(txDeps []sdkacltypes.AccessOperation, _ Tx, _ AnteDepGenerator) ([]sdkacltypes.AccessOperation, error) {
-	return txDeps, nil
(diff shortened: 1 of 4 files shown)


types/handler.go
```diff
@@ -153,5 +153,11 @@ func (t Terminator) AnteHandle(ctx Context, _ Tx, _ bool, _ AnteHandler) (Contex
 
 // Simply return provided txDeps and nil error
 func (t Terminator) AnteDeps(txDeps []sdkacltypes.AccessOperation, _ Tx, _ AnteDepGenerator) ([]sdkacltypes.AccessOperation, error) {
-	return txDeps, nil
+	// TODO:: use the commit access helper - blocked by circular dep. Should move it to the type/accesscontrol instead
+	commit := sdkacltypes.AccessOperation{
+		ResourceType: sdkacltypes.ResourceType_ANY,
```
codchen, line 158: can you help me understand why we need this if ante handler will eventually be followed by the actual msg's dependency list which already ends with a COMMIT?

codchen (approved):

types/handler.go
BrandonWeng: actually, yeah that's a good point. Let me remove this and validate the access ops another way - the validate access ops call always assert that there's a COMMIT at the end, this breaks down when we're just validating the ante handlers

PR #83, opened 2022-11-01
Add DEX resource types and Enable Validation 

BrandonWeng:
## Describe your changes and provide context
Adds more DEX pipes 

Also enables validation checking for ante handler and msgs, actually forgot to uncomment that... 

Also, update the error message to be a little more clear about which access ops are invalid 

## Testing performed to validate your change
Ran e2e on loadtest 
image

PR #86, opened 2022-11-02
Missed nit comments

BrandonWeng:
Missed nit comments from this PR: https://github.com/sei-protocol/sei-cosmos/pull/83


types/handler.go
```diff
@@ -153,9 +153,5 @@ func (t Terminator) AnteHandle(ctx Context, _ Tx, _ bool, _ AnteHandler) (Contex
 
 // Simply return provided txDeps and nil error
 func (t Terminator) AnteDeps(txDeps []sdkacltypes.AccessOperation, _ Tx, _ AnteDepGenerator) ([]sdkacltypes.AccessOperation, error) {
-	txDeps = append(txDeps,
-		sdkacltypes.AccessOperation{ResourceType: sdkacltypes.ResourceType_ANY, AccessType: sdkacltypes.AccessType_COMMIT, IdentifierTemplate: "*"},
-	)
-
```
BrandonWeng, line 159: Don't think we need this, added it previously for the missing messages but it doesnt make a difference. The bug turned out to be in the iteration of the mappings

udpatil (approved):

PR #88, opened 2022-11-15
Add hard rollback support and fix existing rollback 

BrandonWeng:
Add support for hard rollbacks and fixes existing rollback mechanism to actually rollback the CMS 

Cherry-picks this commit: https://github.com/Wryhder/cosmos-sdk/commit/c8142f02e43f062e44d1a6132d557198da272af6 

Sent transactions then rollback back blocks
image

Stopped all validators and hard rolled back, restarted them and saw that they started from the block we rolled back to and then did another run of TXs

### DIFF
--- a/store/rootmulti/store.go
+++ b/store/rootmulti/store.go
@@ -917,29 +917,40 @@ func (rs *Store) buildCommitInfo(version int64) *types.CommitInfo {
 }
 
 // RollbackToVersion delete the versions after `target` and update the latest version.
-func (rs *Store) RollbackToVersion(target int64) int64 {
-	if target < 0 {
-		panic("Negative rollback target")
-	}
-	current := getLatestVersion(rs.db)
-	if target >= current {
-		return current
-	}
-	for ; current > target; current-- {
-		rs.pruneHeights = append(rs.pruneHeights, current)
-	}
-	rs.pruneStores()
-
-	// update latest height
-	bz, err := gogotypes.StdInt64Marshal(current)
-	if err != nil {
-		panic(err)
+func (rs *Store) RollbackToVersion(target int64) error {
+	if target <= 0 {
+		return fmt.Errorf("invalid rollback height target: %d", target)
 	}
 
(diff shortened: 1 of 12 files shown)


store/rootmulti/store.go
```diff
@@ -917,29 +917,40 @@ func (rs *Store) buildCommitInfo(version int64) *types.CommitInfo {
...
+		return fmt.Errorf("invalid rollback height target: %d", target)
 	}
-	current := getLatestVersion(rs.db)
-	if target >= current {
-		return current
-	}
-	for ; current > target; current-- {
-		rs.pruneHeights = append(rs.pruneHeights, current)
-	}
-	rs.pruneStores()
```
philipsu522, line 931: do we not need to prune stores anymore? Seems like the new behavior doesn't do this

BrandonWeng: This was also removed in the original commit 
https://github.com/Wryhder/cosmos-sdk/commit/c8142f02e43f062e44d1a6132d557198da272af6#diff-2bc7a86169d4c2f891d7ae74b21ca4bcc5787d10114e2eee215fe1ddba5c999eL947

BrandonWeng: Also, the behavior for `LoadVersionForOverwriting` is pretty much the same as the pruning method 

https://github.com/cosmos/iavl/blob/c64ca70ffa771de329b0e1a460594e57c4e03555/mutable_tree.go#L618-L646 __

```
	for key, store := range rs.stores {
		if store.GetStoreType() == types.StoreTypeIAVL {
			// If the store is wrapped with an inter-block cache, we must first unwrap
			// it to get the underlying IAVL store.
			store = rs.GetCommitKVStore(key)

			if err := store.(*iavl.Store).DeleteVersions(rs.pruneHeights...); err != nil {
				if errCause := errors.Cause(err); errCause != nil && errCause != iavltree.ErrVersionDoesNotExist {
					panic(err)
				}
			}
		}
	}

```

philipsu522 (approved):

PR #93, opened 2022-11-17
zero div protection

udpatil:
This fixes the zero division issue.

### DIFF
--- a/x/auth/ante/validator_tx_fee.go
+++ b/x/auth/ante/validator_tx_fee.go
@@ -39,6 +39,8 @@ func CheckTxFeeWithValidatorMinGasPrices(ctx sdk.Context, tx sdk.Tx) (sdk.Coins,
 			}
 		}
 	}
+	// this is the lowest priority, and will be used specifically if gas limit is set to 0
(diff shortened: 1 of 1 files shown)


x/auth/ante/validator_tx_fee.go
```diff
@@ -39,8 +39,10 @@ func CheckTxFeeWithValidatorMinGasPrices(ctx sdk.Context, tx sdk.Tx) (sdk.Coins,
 			}
 		}
 	}
-
-	priority := getTxPriority(feeCoins, int64(gas))
+	priority := int64(0)
```
BrandonWeng, line 42: is 0 the lowest priority? might be useful to add a comment here for future readers

BrandonWeng (approved):
lgtm, with one nit

x/auth/ante/validator_tx_fee.go
udpatil: yeah it is, I'll leave a comment. Realistically, if a tx has specified 0 gas, it will most likely run out of gas anyways

PR #96, opened 2022-11-25
Add concurrent cache kv

BrandonWeng:
Copy and pasted the existing cacheKV implementation but changed the maps to use sync map instead and updated the various accessors 

majority of changes in store/concurrentcachekv/store.go
unit test behavior should remain unchanged 

I'll test e2e with the dex change in sei-chain. Would like feedback on this first to make sure we're on the right track


codchen (approved):
thanks for adding this

PR #99, opened 2022-11-29
Add limit to cachekv store

codchen:
To prevent the process from OOMing on large amount of data read (e.g. for order book matching), we will introduce a limit to the cachekv store and its thread-safe equivalent, such that some data will be evicted from the cache if it's full and needs to load something new.

The eviction logic will be random instead of LRU to avoid any overhead that might hurt performance when the cache is not populated fully, which should be the state a validator tries to keep. The tradeoff is that if the limit is reached, a naive random eviction logic will likely incur more evictions than a LRU, so a validator should increase the RAM size and bump the limit soon after they see the limit is reached.

unit test

### DIFF
--- a/store/cachekv/store.go
+++ b/store/cachekv/store.go
@@ -18,37 +18,59 @@ import (
 	dbm "github.com/tendermint/tm-db"
 )
 
-// If value is nil but deleted is false, it means the parent doesn't have the
-// key.  (No need to delete upon Write())
-type cValue struct {
-	value []byte
-	dirty bool
+type mapCacheBackend struct {
+	m map[string]*types.CValue
+}
+
+func (b mapCacheBackend) Get(key string) (val *types.CValue, ok bool) {
+	val, ok = b.m[key]
+	return
+}
+
+func (b mapCacheBackend) Set(key string, val *types.CValue) {
+	b.m[key] = val
+}
+
+func (b mapCacheBackend) Len() int {
+	return len(b.m)
+}
+
+func (b mapCacheBackend) Delete(key string) {
+	delete(b.m, key)
+}
+
+func (b mapCacheBackend) Range(f func(string, *types.CValue) bool) {
+	for k, v := range b.m {
+		if !f(k, v) {
+			break
+		}
+	}
 }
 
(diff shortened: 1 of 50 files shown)


store/cachekv/store.go
```diff
@@ -18,37 +18,59 @@ import (
...
+func (b mapCacheBackend) Len() int {
+	return len(b.m)
+}
+
+func (b mapCacheBackend) Delete(key string) {
+	delete(b.m, key)
+}
+
+func (b mapCacheBackend) Range(f func(string, *types.CValue) bool) {
+	for k, v := range b.m {
```
philipsu522, line 43: should we sort this?

store/concurrentcachekv/store.go
```diff
@@ -20,35 +19,64 @@ import (
...
+	})
+	return
+}
+
+func (b syncMapCacheBackend) Delete(key string) {
+	b.m.Delete(key)
+}
+
+func (b syncMapCacheBackend) Range(f func(string, *types.CValue) bool) {
+	b.m.Range(func(k, v any) bool {
```
philipsu522, line 52: same - should sort

store/types/cache.go
```diff
@@ -0,0 +1,100 @@
...
+		keysToEvict := []string{}
+		c.CacheBackend.Range(func(key string, _ *CValue) bool {
+			keysToEvict = append(keysToEvict, key)
+			len--
+			return len >= c.limit
+		})
+		for _, key := range keysToEvict {
+			c.CacheBackend.Delete(key)
+		}
+	}
```
philipsu522, line 74: we should probably guarantee that the evicted keys is deterministic across vals

codchen: the ordering of the range should now be deterministic given the two sorting done above. However the eviction might still not be deterministic across validators since the total ordering of insertions could be different due to parallel processing (simplest case: cache size 1, val1 insert A first then B whereas val2 B first then A). This should be fine though since this is just cache not something we commit

BrandonWeng (approved):

philipsu522 (approved):

PR #103, opened 2022-12-02
Add max voting power to sei-cosmos

LCyson:
This PR adds a max voting power ratio check for each token delegation. This check will only apply when the bonded voting power > certain threshold (MaxVotingPowerEnforcementThreshold in the params). Both the threshold and max ratio are adjustable as params' fields.

Unit tests


types/staking.go
```diff
@@ -6,6 +6,12 @@ const (
 	// default bond denomination
 	DefaultBondDenom = "stake"
 
+	// default maximal voting power ratio
+	DefaultMaxVotingPowerRatio = "0.05"
```
philipsu522, line 10: does this mean a validator can only have 5% of total delegation?

x/staking/types/params.go
```diff
@@ -179,6 +194,41 @@ func validateMaxEntries(i interface{}) error {
...
+func validateMaxVotingPowerRatio(i interface{}) error {
+	v, ok := i.(sdk.Dec)
+	if !ok {
+		return fmt.Errorf("invalid parameter type: %T", i)
+	}
+
+	if v.IsNil() {
+		return fmt.Errorf("max voting power must be not nil")
+	}
+	if v.IsNegative() {
```
philipsu522, line 222: let's add or if == 0

types/staking.go
philipsu522: if so, do you think we should set something higher?

LCyson: sg will change it to 10%, also this is the default value at genesis and it can be configured later with a gov proposal

x/staking/types/params.go
LCyson: added

philipsu522 (approved):

BrandonWeng (approved):

PR #107, opened 2022-12-05
[baseapp] Bubble up events when concurrent fails for inspection by app

udpatil:
This bubbles up the events from deliverTx when we have a concurrent execution error. This way the app can extract the contract addresses that had the error and appropriately reset their dependencies.

One potential issue that this doesn't resolve is a malicious contract with dependencies intentionally improperly defined calling another contract to force its dependencies to be reset. We can look into this a bit more to identify a solution, but this is unlikely given that a contract needs governance to parallelize its dependencies, meaning that a malicious contract can be socially gatekept in this manner.

### DIFF
--- a/baseapp/abci.go
+++ b/baseapp/abci.go
@@ -244,10 +244,14 @@ func (app *BaseApp) DeliverTx(ctx sdk.Context, req abci.RequestDeliverTx) (res a
 		telemetry.SetGauge(float32(gInfo.GasWanted), "tx", "gas", "wanted")
 	}()
 
-	gInfo, result, _, _, err := app.runTx(ctx.WithTxBytes(req.Tx).WithVoteInfos(app.voteInfos), runTxModeDeliver, req.Tx)
+	gInfo, result, anteEvents, _, err := app.runTx(ctx.WithTxBytes(req.Tx).WithVoteInfos(app.voteInfos), runTxModeDeliver, req.Tx)
 	if err != nil {
 		resultStr = "failed"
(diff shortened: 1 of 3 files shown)


baseapp/abci.go
```diff
@@ -247,6 +247,10 @@ func (app *BaseApp) DeliverTx(ctx sdk.Context, req abci.RequestDeliverTx) (res a
 	gInfo, result, anteEvents, _, err := app.runTx(ctx.WithTxBytes(req.Tx).WithVoteInfos(app.voteInfos), runTxModeDeliver, req.Tx)
 	if err != nil {
 		resultStr = "failed"
+		// if we have a result, use those events instead of just the anteEvents
+		if result != nil {
+			return sdkerrors.ResponseDeliverTxWithEvents(err, gInfo.GasWanted, gInfo.GasUsed, sdk.MarkEventsToIndex(result.Events, app.indexEvents), app.trace)
+		}
```
BrandonWeng, line 253: Are we throwing away the ante events here?

udpatil: no, because the ante events are prepended onto the result events now if the result is non-nil (https://github.com/sei-protocol/sei-cosmos/pull/107/files/af11b782ed830189a7d07e944a69804857d06add#diff-328a55b5ac5816102544792c5c6a75df50df3b9b01afe770c9abc2da79b3e6afR860-R862)

codchen (approved):

BrandonWeng (approved):

PR #110, opened 2022-12-09
[acl] Add more selector and resource types

udpatil:
This adds more selector and resource types for use with wasm depedendencies. Additionally, it fixes up the protocgen.sh script to work properly (with buf version 0.56.0)
Local sei chain testing and existing tests


scripts/protocgen.sh
```diff
@@ -26,12 +26,12 @@ Mgoogle/protobuf/any.proto=github.com/cosmos/cosmos-sdk/codec/types:. \
...
-  -I "third_party/proto" \
-  --doc_out=./docs/core \
-  --doc_opt=./docs/protodoc-markdown.tmpl,proto-docs.md \
-  $(find "$(pwd)/proto" -maxdepth 5 -name '*.proto')
+# buf protoc \
+#   -I "proto" \
+#   -I "third_party/proto" \
+#   --doc_out=./docs/core \
+#   --doc_opt=./docs/protodoc-markdown.tmpl,proto-docs.md \
+#   $(find "$(pwd)/proto" -maxdepth 5 -name '*.proto')
```
philipsu522, line 34: shoudl we undo this? or is the fix here to remove this completely?

x/accesscontrol/keeper/keeper.go
```diff
@@ -226,6 +226,25 @@ func BuildSelectorOps(accessOps []acltypes.AccessOperationWithSelector, senderBe
...
+		case acltypes.AccessOperationSelectorType_CONTRACT_ADDRESS:
+			contractAddress, err := sdk.AccAddressFromBech32(opWithSelector.Selector)
+			if err != nil {
+				return []acltypes.AccessOperationWithSelector{}, err
+			}
+			opWithSelector.Operation.IdentifierTemplate = fmt.Sprintf(
+				opWithSelector.Operation.IdentifierTemplate,
+				hex.EncodeToString(contractAddress),
+			)
+		case acltypes.AccessOperationSelectorType_JQ_MESSAGE_CONDITIONAL:
```
philipsu522, line 238: out of curiosity, what is an example of a jq use case?

philipsu522 (approved):

scripts/protocgen.sh
udpatil: I'll confirm what the fix is and lyk

x/accesscontrol/keeper/keeper.go
udpatil: if for example a contract needs more general resource access across all bank balances (because its send funds to a bunch of people), but only on a execute Msg used infrequently ("send_mass_funds" for example), we can use this message conditional with the format `".send_mass_funds"` so that it ONLY matches if thats the excute command being used, and in that case would invoke the widespread resource access.

TLDR, it mitigates widespread resource access use cases to the execute functions that use them.

PR #115, opened 2022-12-19
Introduce Expedited Proposals 

yzang2019:
## Problem Statement
Currently the long voting window is used to provide opportunities for users to review the proposals and have enough time to vote appropriately to express their opinions. While this is the expected and desirable behavior, it is sometimes not ideal, for example:
1. When the whole system needs to respond immediately to an ongoing threats or a chain needs to enact an emergency upgrade via a text proposal, but needs to bypass the standard voting period.
2. There are some cases where a gov proposal needs to be passed to mitigate an outage, and if voting windows are really long (say a week), the impact of outage could cause disasters
3. Today when a severe vulnerability is uncovered the validator community can discuss an upcoming update in private and prepare an update but it is difficult to do this in a way that does not leak advanced warning for someone to prepare an attack against the network while this is occurring. As the value of assets increase further it is conceivable that even one of the individuals associated with an existing validator in these private channels may be tempted to leverage information to mount an attack before a mitigation is in place.

## Proposed Concept
This change introduces a new option to create an expedited proposal for emergency purpose. If a proposal is expedited, it uses a shorter **voting period** and a higher **tally threshold** and requires to reach a higher **quorum**. If the higher threshold is not met under the short expedited voting period, the expedited proposal is then converted as if it was a regular proposal and tallied / voted on under regular rules.

The distinction is achieved by having an IsExpedited flag on the proposal struct. When a proposal is converted from expedited to regular, this flag is flipped and VotingEndTime is extended. The rest of the fields, including the ProposalId remain the same. As a result, all existing votes are preserved.

The expedited governance proposal option would add a capability for a network to respond rapidly during emergency situations while balancing protection of rights against abuse where a proposal is pushed through quickly without time for proper debate. This can be achieved by providing an additional governance controlled set of parameters that establish a much higher level of approval such as a 67% majority or similar value appropriately chosen by the network.

Similar concept has been proposed and achieved [here](https://wallet.keplr.app/chains/osmosis/proposals/278)

## Changes
- add **IsExpedited** flag to protos and some proposal content struct 
- add **ExpeditedVotingPeriod** to protos of gov voting parameters (Default: 1 day)
- add **ExpeditedThreshold** to protos of gov tally parameters (Default: 2/3)
- add **ExpeditedQuorum** to protos of gov tally parameters (Default: 2/3)
- add **MinExpeditedDeposit** flag to indicate the minimum deposit needed for an expedited proposal to enter voting period
- implement changes in abci of x/gov to move expedited proposals to regular once voting period ends with proposal not meeting ExpeditedThreshold
- update documentation for proto change

**Supporting Proposal:**
- Text Proposal
- Parameter Change Proposal
- Upgrade Proposal

Added unit and integration tests to cover all new logic related to expedited proposals

Integration Test Locally:
1. Start a chain
2. Submit an expedited proposal
3. Make sure Voting End Time should be calculated based on expedited voting period
4. Vote Yes before expedited voting period
5. Proposal should pass
6. Submit another expedited proposal
7. Do not vote and let it expire
8. Make sure Voting End Time is updated based on regular voting period


x/gov/abci.go
```diff
@@ -49,12 +49,20 @@ func EndBlocker(ctx sdk.Context, keeper keeper.Keeper) {
...
+		// EXCEPT when an expedited proposal fails.
+		if !(proposal.IsExpedited && !passes) {
+			if burnDeposits {
+				keeper.DeleteDeposits(ctx, proposal.ProposalId)
+			} else {
+				keeper.RefundDeposits(ctx, proposal.ProposalId)
+			}
 		}
 
+		keeper.RemoveFromActiveProposalQueue(ctx, proposal.ProposalId, proposal.VotingEndTime)
```
philipsu522, line 64: should this logic be here? won't we end up removing active proposals from the queue as well?

philipsu522 (commented):
This change looks really good, especially appreciate the unit tests! One small comment re. removing proposal from queue. 
I think one thing we're missing is the "gov params" migration. Because we're introducing a new field, we'll need to migrate the keeper store params. Here is an example: https://github.com/sei-protocol/sei-chain/blob/master/x/mint/keeper/migrations.go#L18
I believe gov is on version "2", so we'd implement a similar Migrate2To3 function

x/gov/abci.go
LCyson: +1

LCyson (commented):
can we use an independent expediatedDeposit amount in the gov param, so that we can use expensive deposit fee to prevent people from abusing the expedited proposals

LCyson:
can we also test if normal proposal works properly in the e2e test

x/gov/abci.go
yzang2019: The original logic is to remove it anyway. I think we move it here because we will have to add it back later if this is an expedited proposal, then it would become a regular one.

yzang2019:
> can we use an independent expediatedDeposit amount in the gov param, so that we can use expensive deposit fee to prevent people from abusing the expedited proposals

Yeah, that's a great idea! I'll add that next and also perform more integration tests. Thanks for review.

philipsu522 (approved):

yzang2019:
@philipsu522 Given we have other high priorities, and this change is already pretty huge, does it make sense to do the migration change in a separate PR later?

yzang2019:
> can we also test if normal proposal works properly in the e2e test

We've done e2e test in a 4 node test cluster:
- proposal
- expedited proposal
- expedited proposal fallback to regular proposal
- validate new parameter minExpeditedDeposit is working
- validate new parameter expeditedVotingPeriod is working
- validate new parameter expeditedThreshold is working
- validate new parameter expeditedQuorum is working

yzang2019:
> can we use an independent expediatedDeposit amount in the gov param, so that we can use expensive deposit fee to prevent people from abusing the expedited proposals

Added

x/gov/abci.go
yzang2019: It was move from line 93 to here. The reason is because we need to add it back on line 104

LCyson (approved):

PR #118, opened 2022-12-21
Add midblock to sei-cosmos

udpatil:
This adds the midblock interfaces and module manager functionality to sei-cosmos

Added unit test for module manager behavior and also tested with local sei with sei-chain changes

### DIFF
--- a/baseapp/abci.go
+++ b/baseapp/abci.go
@@ -160,6 +160,24 @@ func (app *BaseApp) BeginBlock(ctx sdk.Context, req abci.RequestBeginBlock) (res
 	return res
 }
 
+func (app *BaseApp) MidBlock(ctx sdk.Context, height int64) (events []abci.Event) {
+	defer telemetry.MeasureSince(time.Now(), "abci", "mid_block")
+
+	if app.midBlocker != nil {
+		midBlockEvents := app.midBlocker(ctx, height)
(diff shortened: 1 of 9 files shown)


baseapp/abci.go
```diff
@@ -160,6 +160,24 @@ func (app *BaseApp) BeginBlock(ctx sdk.Context, req abci.RequestBeginBlock) (res
 	return res
 }
 
+func (app *BaseApp) MidBlock(ctx sdk.Context, height int64) (events []abci.Event) {
+	defer telemetry.MeasureSince(time.Now(), "abci", "mid_block")
```
BrandonWeng, line 164: Should we also add tracing to midblock?

udpatil: can you link an example of the tracing? I just included the same telemetry reporting as was present for endblock since midblock is fairly similar

codchen: tracing rn is only set on sei-chain level i think. To add it for MidBlock you can do it here https://github.com/sei-protocol/sei-chain/blob/master/app/abci.go#L24

codchen (approved):

baseapp/abci.go
udpatil: ah gotcha thanks, i've included the tracing in the sei-chain PR

PR #119, opened 2022-12-21
Fix Hard Rollback for App state 

BrandonWeng:
Previous hard rollback implementation was based on the OSS code, but it doesn't seem to be working properly if we tried to roll back a block with any state we run into this error. Upon investigation, before the fix, we're still able to query the hard "deleted" block with the AVL tool so it doesn't seem to be properly deleted. 
 * to fix this, we need to actually commit the change to disk after the load version 

```
4:31PM INF stopping service module=consensus service=Group wal=/Users/brandon/.sei/data/cs.wal/wal
panic: version 523 was already saved to different hash 7F364931ECF614D736CD4AB638107E2A9CD9A17ACF0CC3B1027EFE830BD17C33 (existing hash D1F0D96E6376D254072A840755B6B5E796086A5C648F5E65E900F60CB31345EA) [recovered]
        panic: version 523 was already saved to different hash 7F364931ECF614D736CD4AB638107E2A9CD9A17ACF0CC3B1027EFE830BD17C33 (existing hash D1F0D96E6376D254072A840755B6B5E796086A5C648F5E65E900F60CB31345EA) [recovered]
        panic: version 523 was already saved to different hash 7F364931ECF614D736CD4AB638107E2A9CD9A17ACF0CC3B1027EFE830BD17C33 (existing hash D1F0D96E6376D254072A840755B6B5E796086A5C648F5E65E900F60CB31345EA) [recovered]
        panic: version 523 was already saved to different hash 7F364931ECF614D736CD4AB638107E2A9CD9A17ACF0CC3B1027EFE830BD17C33 (existing hash D1F0D96E6376D254072A840755B6B5E796086A5C648F5E65E900F60CB31345EA) [recovered]
        panic: version 523 was already saved to different hash 7F364931ECF614D736CD4AB638107E2A9CD9A17ACF0CC3B1027EFE830BD17C33 (existing hash D1F0D96E6376D254072A840755B6B5E796086A5C648F5E65E900F60CB31345EA) [recovered]
        panic: version 523 was already saved to different hash 7F364931ECF614D736CD4AB638107E2A9CD9A17ACF0CC3B1027EFE830BD17C33 (existing hash D1F0D96E6376D254072A840755B6B5E796086A5C648F5E65E900F60CB31345EA) [recovered]
        panic: version 523 was already saved to different hash 7F364931ECF614D736CD4AB638107E2A9CD9A17ACF0CC3B1027EFE830BD17C33 (existing hash D1F0D96E6376D254072A840755B6B5E796086A5C648F5E65E900F60CB31345EA) [recovered]
        panic: version 523 was already saved to different hash 7F364931ECF614D736CD4AB638107E2A9CD9A17ACF0CC3B1027EFE830BD17C33 (existing hash D1F0D96E6376D254072A840755B6B5E796086A5C648F5E65E900F60CB31345EA)
```

Blocks with TXs:
image

Rolled back
image

See that blocks didn't have any TXs 
image

Ran it again with hard rollback and see that it works fine as expected


store/rootmulti/store.go
```diff
@@ -1131,19 +1113,6 @@ func getPruningHeights(db dbm.DB) ([]int64, error) {
 	return prunedHeights, nil
 }
 
-func flushMetadata(db dbm.DB, version int64, cInfo *types.CommitInfo, pruneHeights []int64) {
```
BrandonWeng, line 1134: Removing this to consolidate, right now we have two flushMetaData functions

store/rootmulti/store.go
```diff
@@ -1162,3 +1131,14 @@ func flushLatestVersion(batch dbm.Batch, version int64) {
 
 	batch.Set([]byte(latestVersionKey), bz)
 }
+
+func flushPruningHeights(batch dbm.Batch, pruneHeights []int64) {
```
BrandonWeng, line 1135: pretty much the same "setPruningHeights" function, just more inline with the commit info and flush latest version

store/rootmulti/store.go
```diff
@@ -978,23 +990,24 @@ func (rs *Store) RollbackToVersion(target int64) error {
 			}
 		}
 	}
-
-	rs.flushMetadata(rs.db, target, rs.buildCommitInfo(target))
+	rs.lastCommitInfo = commitStores(target, rs.stores)
```
BrandonWeng, line 993: I think we need to commit the store when rolling back, or else the height is still queriable and if we try to rollback where there's state it'll complain that the app hash was already committed with a different state

store/rootmulti/store.go
```diff
@@ -73,6 +73,19 @@ var (
...
+	for key := range m {
+		keys = append(keys, key)
+	}
+	sort.Slice(keys, func(i, j int) bool {
+		ki, kj := keys[i], keys[j]
+		return ki.Name() < kj.Name()
+	})
+	return keys
+}
+
```
BrandonWeng, line 88: this is from latest cosmos

server/rollback.go
```diff
@@ -40,16 +40,38 @@ application.
 				ctx.Viper,
 			)
 
+			// App State
+			lastCommit := app.CommitMultiStore().LastCommitID()
+			fmt.Printf("Initial App state height=%d and hash=%X\n", lastCommit.GetVersion(), lastCommit.GetHash())
```
BrandonWeng, line 45: Adding more print statements, when rolling back we should be a bit more verbose to understand the before an after pictures

server/rollback.go
```diff
@@ -40,16 +40,38 @@ application.
...
+			fmt.Printf("Rolled back app state to height %d and hash %X\n", lastCommit.GetVersion(), lastCommit.GetHash())
+
+			// This will cause issues when you try starting the chain. Something probably went wrong
+			if removeBlock && app.CommitMultiStore().LastCommitID().Version != tmHeight {
+				panic("Application state height does not match the tendermint state height")
+			}
+
+			if removeBlock && string(lastCommit.GetHash()) != string(hash) {
+				panic("Application state hash does not match the tendermint state hash")
+			}
```
BrandonWeng, line 72: These are checked in the handshake anyways, would help catch things earlier

philipsu522 (approved):

PR #123, opened 2023-01-04
Parse is_expedited json field

BrandonWeng:
is_expedited wasn't being parsed properly from the files 

```
{
    "title": "Test Proposal",
    "description": "My awesome proposal",
    "is_expedited": true,
    "type": "Text",
    "deposit": "1000000000usei"
  }
```

```
 seid tx gov submit-proposal --proposal=./test.json --from admin  -y  --broadcast-mode=block --fees 1000sei
```

image

### DIFF
--- a/x/gov/client/cli/tx.go
+++ b/x/gov/client/cli/tx.go
@@ -30,7 +30,7 @@ const (
 )
 
 type proposal struct {
-	Title       string ` json:"title"`
+	Title       string  `json:"title"`
 	Description string  `json:"description"`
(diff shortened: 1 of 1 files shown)


x/gov/client/cli/tx.go
```diff
@@ -33,7 +33,7 @@ type proposal struct {
 	Title       string
```
yzang2019, line 33: since we are adding json annotation, can we add this for all other fields?

yzang2019 (approved):

PR #126, opened 2023-01-06
[acl] Add selector type and handler for generating dependencies from contract ref

udpatil:
This adds a selector type for generating dependencies based on another contract's registered dependencies.

Added unit tests for generating contract dependencies and also for cases where the contract B dependencies may not exist in which case we fall back to using synchronous dependencies


x/accesscontrol/keeper/keeper.go
```diff
@@ -250,6 +250,26 @@ func BuildSelectorOps(accessOps []acltypes.AccessOperationWithSelector, senderBe
...
+			// for now, we will just pass in the same message body, this needs to be changed later though
+			// TODO: build new msgbody for the new contract execute / query msg in later milestone tasks
+			wasmDeps, err := k.GetWasmDependencyMapping(ctx, interContractAddress, contractAddr.String(), msgBody, true)
+
+			if err != nil || !wasmDeps.Enabled {
+				// if we have an error fetching the dependency mapping or the mapping is disabled, we want to use the synchronous mappings instead
+				selectedAccessOps = append(selectedAccessOps, types.SynchronousAccessOpsWithSelector()...)
+			} else {
+				// if we did get deps properly and they are enabled, now we want to add them to our access operations
+				selectedAccessOps = append(selectedAccessOps, wasmDeps.AccessOps...)
```
LCyson, line 269: will this significantly increase the size of accessOps list, suppose there are a contract depends on 100 other contracts then it will have an accessOps list 100x bigger than the others. Maybe we can use a Map instead of List in the future?

udpatil: this is true, we can apply a step of deduplicating identical access operations in the future, but I don't think its a significant issue right now. Good call out though, I'lll create a task for the future to do this

x/accesscontrol/keeper/keeper.go
```diff
@@ -250,6 +250,26 @@ func BuildSelectorOps(accessOps []acltypes.AccessOperationWithSelector, senderBe
...
+				return []acltypes.AccessOperationWithSelector{}, err
+			}
+			// TODO: add a circular dependency check here to ignore if we've already seen this contract/identifier in our reference chain
+			// for now, we will just pass in the same message body, this needs to be changed later though
+			// TODO: build new msgbody for the new contract execute / query msg in later milestone tasks
+			wasmDeps, err := k.GetWasmDependencyMapping(ctx, interContractAddress, contractAddr.String(), msgBody, true)
+
+			if err != nil || !wasmDeps.Enabled {
+				// if we have an error fetching the dependency mapping or the mapping is disabled, we want to use the synchronous mappings instead
+				selectedAccessOps = append(selectedAccessOps, types.SynchronousAccessOpsWithSelector()...)
```
codchen, line 266: i think we can just set `selectedAccessOps = SynchronousAccessOpsWithSelector` in this case?

codchen (approved):

x/accesscontrol/keeper/keeper.go
```diff
@@ -250,6 +250,26 @@ func BuildSelectorOps(accessOps []acltypes.AccessOperationWithSelector, senderBe
...
 			)
+		case acltypes.AccessOperationSelectorType_CONTRACT_REFERENCE:
+			// We use this to import the dependencies from another contract address
+			interContractAddress, err := sdk.AccAddressFromBech32(opWithSelector.Selector)
+			if err != nil {
+				return []acltypes.AccessOperationWithSelector{}, err
+			}
+			// TODO: add a circular dependency check here to ignore if we've already seen this contract/identifier in our reference chain
+			// for now, we will just pass in the same message body, this needs to be changed later though
+			// TODO: build new msgbody for the new contract execute / query msg in later milestone tasks
```
codchen, line 261: we might want to add another TODO so that reference selectors of the referred contract are recursively evaluated

x/accesscontrol/keeper/keeper.go
LCyson: sg can we add a TODO here

LCyson (approved):

x/accesscontrol/keeper/keeper.go
udpatil: This is already done since we call `GetWasmDependencyMapping` from within build selectors, but might be worth adding a test for this.

x/accesscontrol/keeper/keeper.go
udpatil: I think we would like to append, since previous access operations with their selectors are also contained here within selectedAccessOps

udpatil: done

x/accesscontrol/keeper/keeper.go
udpatil: done

PR #128, opened 2023-01-10
[ACL] modify WASM dependency proto for partitioning

codchen:
This PR replaces `WasmDependencyMapping`'s old `access_ops` field with three new fields:
- `base_access_ops`: a list of access operations common to all message types
- `query_access_ops`: a list of lists of access operations per query message type
- `execute_access_ops`: a list of lists of access operations per execute message type

This PR also made two other changes:
- Added a separate `GetRawWasmDependencyMapping` method for retrieving the proto value as-is without populating any data. The original `GetWasmDependencyMapping` now always populate data and return just the populated access op list.
- Added a access op merging utility that 1. reduce the list to default synchronous ops if either side if synchronous 2. dedupe COMMIT ops
unit tests


proto/cosmos/accesscontrol/accesscontrol.proto
```diff
@@ -28,9 +33,9 @@ message MessageDependencyMapping {
 
 message WasmDependencyMapping {
     bool enabled = 1;
```
udpatil, line 35: I think we can probably deprecate the enabled actually, since we don't meaningfully use it and maintaining behavior around it is additional effort that's not necesarily worth it. The main use case would be for resetting wasm dependencies, but we've been leaving it enabled but with synchronous dependencies instead. What we could do instead is simply reset it to synchronous or even delete the wasm dependency to "reset" it aka clear out unneeded store state and rely on the behavior that returns Synchronous access ops when no registered dependencies exist.
wdyt @codchen

x/accesscontrol/keeper/keeper.go
```diff
@@ -125,48 +126,60 @@ func (k Keeper) SetDependencyMappingDynamicFlag(ctx sdk.Context, messageKey type
...
-	return dependencyMapping, nil
+
+	// use synchronous if dependency mapping is disabled
+	return types.SynchronousWasmAccessOps(), nil
 }
 
-func (k Keeper) BuildSelectorOps(ctx sdk.Context, contractAddr sdk.AccAddress, accessOps []acltypes.AccessOperationWithSelector, senderBech string, msgBody []byte, circularDepLookup ContractReferenceLookupMap) ([]acltypes.AccessOperationWithSelector, error) {
-	selectedAccessOps := []acltypes.AccessOperationWithSelector{}
+func (k Keeper) BuildSelectorOps(ctx sdk.Context, contractAddr sdk.AccAddress, accessOps []*acltypes.WasmAccessOperation, senderBech string, msgBody []byte, circularDepLookup ContractReferenceLookupMap) ([]*acltypes.WasmAccessOperation, error) {
+	selectedAccessOps := []*acltypes.WasmAccessOperation{}
```
udpatil, line 175: if we're restructuring this data type, should we make it a map[accessOperation]struct{} instead to basically be a set? we previously had a TODO to deduplicate access operations, and this might be a good way to go about it? I'm ok leaving this as is and addressing it later instead for simplicity

x/accesscontrol/keeper/keeper.go
```diff
@@ -317,15 +330,17 @@ func (k Keeper) ResetWasmDependencyMapping(
...
 ) error {
-	dependencyMapping, err := k.GetWasmDependencyMapping(ctx, contractAddress, "", []byte{}, false, make(ContractReferenceLookupMap))
+	dependencyMapping, err := k.GetRawWasmDependencyMapping(ctx, contractAddress)
 	if err != nil {
 		return err
 	}
 	store := ctx.KVStore(k.storeKey)
 	// keep `Enabled` true so that it won't cause all WASM resources to be synchronous
-	dependencyMapping.AccessOps = types.SynchronousAccessOpsWithSelector()
+	dependencyMapping.BaseAccessOps = types.SynchronousWasmAccessOps()
```
udpatil, line 339: maybe we simply delete the key instead such that it returns synchronous by default as the behavior in GetWasmDependencyMapping to return sync access ops if no registered deps exist?

udpatil: I guess it could make sense to keep this to persist the reset reason, but couldn't we expose the reset reason via error logging instead, or is that not good enough since the logs would be volatile instead of persistent?

udpatil:
^ Rebased the change onto me change for the "*" JQ handling

x/accesscontrol/types/message_dependency_mapping.go
```diff
@@ -108,8 +122,25 @@ func DefaultWasmDependencyMappings() []acltypes.WasmDependencyMapping {
 	return []acltypes.WasmDependencyMapping{}
 }
 
+func MergeWasmAccessOps(accessOps1 []*acltypes.WasmAccessOperation, accessOps2 []*acltypes.WasmAccessOperation) []*acltypes.WasmAccessOperation {
+	if accessOps1 == nil || len(accessOps1) == 0 {
```
udpatil, line 126: the nil check isn't necessary right? since we're taking in an array of pointers instead of a pointer to an array of pointers?

proto/cosmos/accesscontrol/accesscontrol.proto
codchen: yeah i agree. Removed

x/accesscontrol/keeper/keeper.go
codchen: I was thinking about letting the contract owner know why their contract gets reset. If the documentation makes that clear enough though I think deleting the key could work (and better)

x/accesscontrol/keeper/keeper.go
codchen: yeah makes sense. Updated

x/accesscontrol/types/message_dependency_mapping.go
codchen: removed this func in favor of the new set usage

x/accesscontrol/keeper/keeper.go
udpatil: yeah, this sounds fine, we can keep it in that case

x/accesscontrol/types/access_operations.go
```diff
@@ -0,0 +1,63 @@
+package types
+
+import (
+	acltypes "github.com/cosmos/cosmos-sdk/types/accesscontrol"
+)
+
+type AccessOperationSet struct {
```
udpatil, line 7: curious, why do we want a struct containing a map instead just having like

```
type AccessOperationSet map[aclypes.AccessOperation]struct{}
```

x/accesscontrol/types/access_operations.go
```diff
@@ -0,0 +1,63 @@
...
+func (waos *AccessOperationSet) HasIdentifier(identifier string) bool {
+	for op := range waos.ops {
+		if op.IdentifierTemplate == identifier {
+			return true
+		}
+	}
+	return false
+}
+
+func (waos *AccessOperationSet) HasResourceType(rt acltypes.ResourceType) bool {
```
udpatil, line 56: is this also intended for testing purposes only?

x/accesscontrol/types/access_operations.go
udpatil, line 7: should we also add a function to convert the map into a list of access operations? I don't think the order matters apart from the fact that the COMMIT access op should be last

x/accesscontrol/keeper/keeper.go
```diff
@@ -125,39 +126,46 @@ func (k Keeper) SetDependencyMappingDynamicFlag(ctx sdk.Context, messageKey type
...
+		return nil, sdkerrors.ErrKeyNotFound
+	}
+	dependencyMapping := acltypes.WasmDependencyMapping{}
+	if err := k.cdc.Unmarshal(b, &dependencyMapping); err != nil {
+		return nil, err
+	}
+	return &dependencyMapping, nil
+}
+
+func (k Keeper) GetWasmDependencyMapping(ctx sdk.Context, contractAddress sdk.AccAddress, senderBech string, msgBody []byte, circularDepLookup ContractReferenceLookupMap) (*types.AccessOperationSet, error) {
```
udpatil, line 142: we still would like to return a slice of access operations since that's what the dependency generator (in sei-chain repo) will pass back to the dependency graph generation logic.

(see comment below on having list conversion helper function)

codchen: updated

x/accesscontrol/types/access_operations.go
codchen: added

x/accesscontrol/types/access_operations.go
codchen: yeah

x/accesscontrol/types/access_operations.go
codchen: to hide the underlying map so that callers can't do things like `del` on it directly

x/accesscontrol/types/access_operations.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+func (waos *AccessOperationSet) ToSlice() []acltypes.AccessOperation {
+	res := []acltypes.AccessOperation{}
+	for op := range waos.ops {
+		if op != *CommitAccessOp() {
+			res = append(res, op)
+		}
+	}
+	res = append(res, *CommitAccessOp())
```
udpatil, line 53: it's not guaranteed to have a commit access op right? its probably fine to always append but is there any scenario where we would intend to not append the commit IF it doesnt exist in the map?

x/accesscontrol/keeper/keeper.go
```diff
@@ -125,39 +126,46 @@ func (k Keeper) SetDependencyMappingDynamicFlag(ctx sdk.Context, messageKey type
...
+		return nil, sdkerrors.ErrKeyNotFound
+	}
+	dependencyMapping := acltypes.WasmDependencyMapping{}
+	if err := k.cdc.Unmarshal(b, &dependencyMapping); err != nil {
+		return nil, err
+	}
+	return &dependencyMapping, nil
+}
+
+func (k Keeper) GetWasmDependencyMapping(ctx sdk.Context, contractAddress sdk.AccAddress, senderBech string, msgBody []byte, circularDepLookup ContractReferenceLookupMap) ([]acltypes.AccessOperation, error) {
```
udpatil, line 142: should we rename this to `GetWasmDependencyAccessOps` for more clarity?

udpatil (approved):

x/accesscontrol/types/access_operations.go
codchen: changed so that it only appends if it exists

PR #130, opened 2023-01-11
[acl] pass message info when getting wasm access ops

codchen:
Callers to `GetWasmDependencyAccessOps` now needs to pass in a `WasmMessageInfo` instead of raw message bytes so that the function can identify whether the message is a query or execute. This PR introduced helpers under `types/wasm.go` to help creating such `WasmMessageInfo`. `GetWasmDependencyAccessOps` would then extend the access operations list with the appropriate specific access ops.

Next steps:
- Add separate selector enums for query reference and execute reference
- Handle new selector enums by passing the current WasmMessageInfo

unit tests

### DIFF
--- a/x/accesscontrol/keeper/keeper.go
+++ b/x/accesscontrol/keeper/keeper.go
@@ -158,14 +158,15 @@ func (k Keeper) GetWasmDependencyAccessOps(ctx sdk.Context, contractAddress sdk.
 
 	accessOps := dependencyMapping.BaseAccessOps
 	specificAccessOpsMapping := []*acltypes.WasmAccessOperations{}
-	if msgInfo.MessageType == acltypes.WasmMessageSubtype_EXECUTE && dependencyMapping.ExecuteAccessOps != nil {
+	if msgInfo.MessageType == acltypes.WasmMessageSubtype_EXECUTE && len(dependencyMapping.ExecuteAccessOps) > 0 {
 		specificAccessOpsMapping = dependencyMapping.ExecuteAccessOps
-	} else if msgInfo.MessageType == acltypes.WasmMessageSubtype_QUERY && dependencyMapping.QueryAccessOps != nil {
+	} else if msgInfo.MessageType == acltypes.WasmMessageSubtype_QUERY && len(dependencyMapping.QueryAccessOps) > 0 {
 		specificAccessOpsMapping = dependencyMapping.QueryAccessOps
 	}
 	for _, specificAccessOps := range specificAccessOpsMapping {
 		if specificAccessOps.MessageName == msgInfo.MessageName {
 			accessOps = append(accessOps, specificAccessOps.WasmOperations...)
+			break
 		}
 	}
 
--- a/x/accesscontrol/keeper/keeper.go
+++ b/x/accesscontrol/keeper/keeper.go
@@ -139,7 +139,7 @@ func (k Keeper) GetRawWasmDependencyMapping(ctx sdk.Context, contractAddress sdk
 	return &dependencyMapping, nil
 }
 
-func (k Keeper) GetWasmDependencyAccessOps(ctx sdk.Context, contractAddress sdk.AccAddress, senderBech string, msgBody []byte, circularDepLookup ContractReferenceLookupMap) ([]acltypes.AccessOperation, error) {
+func (k Keeper) GetWasmDependencyAccessOps(ctx sdk.Context, contractAddress sdk.AccAddress, senderBech string, msgInfo *types.WasmMessageInfo, circularDepLookup ContractReferenceLookupMap) ([]acltypes.AccessOperation, error) {
 	uniqueIdentifier := contractAddress.String()
 	if _, ok := circularDepLookup[uniqueIdentifier]; ok {
 		// we've already seen this contract, we should simply return synchronous access Ops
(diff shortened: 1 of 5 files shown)


x/accesscontrol/keeper/keeper.go
```diff
@@ -156,15 +156,27 @@ func (k Keeper) GetWasmDependencyAccessOps(ctx sdk.Context, contractAddress sdk.
...
+	specificAccessOpsMapping := []*acltypes.WasmAccessOperations{}
+	if msgInfo.MessageType == acltypes.WasmMessageSubtype_EXECUTE && dependencyMapping.ExecuteAccessOps != nil {
+		specificAccessOpsMapping = dependencyMapping.ExecuteAccessOps
+	} else if msgInfo.MessageType == acltypes.WasmMessageSubtype_QUERY && dependencyMapping.QueryAccessOps != nil {
+		specificAccessOpsMapping = dependencyMapping.QueryAccessOps
+	}
+	for _, specificAccessOps := range specificAccessOpsMapping {
+		if specificAccessOps.MessageName == msgInfo.MessageName {
+			accessOps = append(accessOps, specificAccessOps.WasmOperations...)
+		}
```
udpatil, line 169: should we break at this point because we only expect 1 entry for a given messageName?

x/accesscontrol/keeper/keeper_test.go
```diff
@@ -122,6 +122,92 @@ func TestWasmDependencyMapping(t *testing.T) {
 	require.Error(t, aclkeeper.ErrWasmDependencyMappingNotFound, err)
 }
 
+func TestWasmDependencyMappingWithExecuteMsgInfo(t *testing.T) {
+	app := simapp.Setup(false)
+	ctx := app.BaseApp.NewContext(false, tmproto.Header{})
+
+	wasmContractAddresses := simapp.AddTestAddrsIncremental(app, ctx, 1, sdk.NewInt(30000000))
+	wasmContractAddress := wasmContractAddresses[0]
+	wasmMapping := acltypes.WasmDependencyMapping{
```
udpatil, line 131: One thing I realized for this new wasmdependencyMapping is that we need to update the validation logic in the message_dependency_mapping.go file. More specifically, do we want  to enforce COMMIT access op ONLY for the base access ops?

x/accesscontrol/keeper/keeper.go
udpatil: additionally, if this is the case, we should update our validation logic for wasm dependency mappings to enforce this.

x/accesscontrol/keeper/keeper.go
```diff
@@ -156,15 +156,27 @@ func (k Keeper) GetWasmDependencyAccessOps(ctx sdk.Context, contractAddress sdk.
 		return nil, err
 	}
 
-	// TODO: extend base access op with message-specific ops once message type identifier is passed in
-	selectedAccessOps, err := k.BuildSelectorOps(ctx, contractAddress, dependencyMapping.BaseAccessOps, senderBech, msgBody, circularDepLookup)
+	accessOps := dependencyMapping.BaseAccessOps
+	specificAccessOpsMapping := []*acltypes.WasmAccessOperations{}
+	if msgInfo.MessageType == acltypes.WasmMessageSubtype_EXECUTE && dependencyMapping.ExecuteAccessOps != nil {
```
udpatil, line 161: this should be a len check, not a nil check right? because we would have a ezro length array instead of a nil pointer right?

x/accesscontrol/keeper/keeper.go
```diff
@@ -156,15 +156,27 @@ func (k Keeper) GetWasmDependencyAccessOps(ctx sdk.Context, contractAddress sdk.
...
+		specificAccessOpsMapping = dependencyMapping.ExecuteAccessOps
+	} else if msgInfo.MessageType == acltypes.WasmMessageSubtype_QUERY && dependencyMapping.QueryAccessOps != nil {
```
udpatil, line 163: same here

x/accesscontrol/types/wasm.go
```diff
@@ -0,0 +1,53 @@
...
+
+func newMessageInfo(fullBody []byte, messageType acltypes.WasmMessageSubtype) (*WasmMessageInfo, error) {
+	name, body, err := extractMessage(fullBody)
+	if err != nil {
+		return nil, err
+	}
+	return &WasmMessageInfo{
+		MessageType:     messageType,
+		MessageName:     name,
+		MessageBody:     body,
```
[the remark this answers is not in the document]
codchen: I think fullbody since they need to use fullbody anyway for BaseAccessOps so might want to keep it consistent

x/accesscontrol/types/wasm.go
```diff
@@ -0,0 +1,53 @@
...
+		MessageFullBody: fullBody,
+	}, nil
+}
+
+// WASM message body is JSON-serialized and use the message name
+// as the only top-level key
+func extractMessage(fullBody []byte) (string, []byte, error) {
+	var deserialized map[string]json.RawMessage
+	if err := json.Unmarshal(fullBody, &deserialized); err != nil {
+		return "", nil, err
```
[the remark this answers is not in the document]
codchen: i changed it to return `""` and `fullBody`, which would cause GetWasmAccessOps to use BaseAccessOperations only, and then the wasm module can detect the malformat body and handle it the same way as before

x/accesscontrol/keeper/keeper_test.go
codchen: yeah that's what we currently do

x/accesscontrol/keeper/keeper.go
codchen: good point. Added validation and break

udpatil (approved):

PR #132, opened 2023-01-11
Add Contract References as a separate struct from Access ops in Wasm Dep Mappings

udpatil:
This separates the inter-contract references from normal access ops for wasm dependencies. It introduces three types of contract references, base, execute, and query contract references. Base contract references are always imported, execute contract refs are imported specifically based on the execute message name, and likewise for query contract references. As an additional note, QUERY contract references cannot contain non-query message types, since queries are read only workflows for cosmwasm contracts.

Updated existing unit tests and adding more shortly

### DIFF
--- a/proto/cosmos/accesscontrol/accesscontrol.proto
+++ b/proto/cosmos/accesscontrol/accesscontrol.proto
@@ -18,6 +18,18 @@ message WasmAccessOperation {
     string selector = 3;
 }
 
+message WasmContractReference {
+    string contract_address = 1;
+    WasmMessageSubtype message_type = 2;
+    string message_name = 3;
+    // TODO: add wasm message adapter format here
+}
+
+message WasmContractReferences {
+    string message_name = 1;
+    repeated WasmContractReference contract_references = 2;
+}
+
 message WasmAccessOperations {
     string message_name = 1;
     repeated WasmAccessOperation wasm_operations = 2;
@@ -35,6 +47,9 @@ message WasmDependencyMapping {
     repeated WasmAccessOperation base_access_ops = 1;
     repeated WasmAccessOperations query_access_ops = 2;
     repeated WasmAccessOperations execute_access_ops = 3;
-    string reset_reason = 4;
-    string contract_address = 5;
+    repeated WasmContractReference base_contract_references = 4;
+    repeated WasmContractReferences query_contract_references = 5;
+    repeated WasmContractReferences execute_contract_references = 6;
+    string reset_reason = 7;
+    string contract_address = 8;
 }
--- a/x/accesscontrol/keeper/keeper.go
+++ b/x/accesscontrol/keeper/keeper.go
@@ -172,7 +172,6 @@ func (k Keeper) GetWasmDependencyAccessOps(ctx sdk.Context, contractAddress sdk.
 
 	selectedAccessOps, err := k.BuildSelectorOps(ctx, contractAddress, accessOps, senderBech, msgInfo, circularDepLookup)
 	if err != nil {
-		// TODO: should we reset wasm dependency mapping here?
 		return nil, err
 	}
 
@@ -188,11 +187,11 @@ func (k Keeper) GetWasmDependencyAccessOps(ctx sdk.Context, contractAddress sdk.
 	for _, specificContractRef := range specificContractRefs {
 		if specificContractRef.MessageName == msgInfo.MessageName {
 			contractRefs = append(contractRefs, specificContractRef.ContractReferences...)
+			break
 		}
 	}
 	importedAccessOps, err := k.ImportContractReferences(ctx, contractAddress, contractRefs, senderBech, msgInfo, circularDepLookup)
 	if err != nil {
-		// TODO: should we reset wasm dependency mapping here?
 		return nil, err
 	}
 	// combine the access ops to get the definitive list of access ops for the contract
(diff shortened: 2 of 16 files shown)


x/accesscontrol/keeper/keeper.go
```diff
@@ -156,21 +156,95 @@ func (k Keeper) GetWasmDependencyAccessOps(ctx sdk.Context, contractAddress sdk.
...
+	}
+	for _, specificAccessOps := range specificAccessOpsMapping {
+		if specificAccessOps.MessageName == msgInfo.MessageName {
+			accessOps = append(accessOps, specificAccessOps.WasmOperations...)
+		}
+	}
+
+	selectedAccessOps, err := k.BuildSelectorOps(ctx, contractAddress, accessOps, senderBech, msgInfo, circularDepLookup)
 	if err != nil {
+		// TODO: should we reset wasm dependency mapping here?
```
udpatil, line 174: cc @codchen do you think we should be resetting the wasm dependencies if we get an error as part of building selector ops or contract references?

udpatil:
Currently setting base branch to tony's change that its based off of so that the changes shown are correct, once that's merged I'll change the base branch to main

x/accesscontrol/keeper/keeper.go
```diff
@@ -171,18 +171,80 @@ func (k Keeper) GetWasmDependencyAccessOps(ctx sdk.Context, contractAddress sdk.
...
 		return nil, err
 	}
-	return selectedAccessOps, nil
+
+	// imports base contract references
+	contractRefs := dependencyMapping.BaseContractReferences
+	// add the specific execute or query contract references based on message type + name
+	specificContractRefs := []*acltypes.WasmContractReferences{}
+	if msgInfo.MessageType == acltypes.WasmMessageSubtype_EXECUTE && dependencyMapping.ExecuteAccessOps != nil {
```
udpatil, line 182: this should be a len check, not a nil check right? because we would have a ezro length array instead of a nil pointer right?

x/accesscontrol/keeper/keeper.go
```diff
@@ -171,18 +171,80 @@ func (k Keeper) GetWasmDependencyAccessOps(ctx sdk.Context, contractAddress sdk.
...
+		specificContractRefs = dependencyMapping.ExecuteContractReferences
+	} else if msgInfo.MessageType == acltypes.WasmMessageSubtype_QUERY && dependencyMapping.QueryAccessOps != nil {
```
udpatil, line 184: same here?

philipsu522 (approved):

x/accesscontrol/keeper/keeper.go
codchen: the error could be from a bad message body rather than a bad dependency registration, in which case we probably don't want to reset

x/accesscontrol/keeper/keeper.go
```diff
@@ -171,18 +171,80 @@ func (k Keeper) GetWasmDependencyAccessOps(ctx sdk.Context, contractAddress sdk.
...
+	// add the specific execute or query contract references based on message type + name
+	specificContractRefs := []*acltypes.WasmContractReferences{}
+	if msgInfo.MessageType == acltypes.WasmMessageSubtype_EXECUTE && len(dependencyMapping.ExecuteContractReferences) != 0 {
+		specificContractRefs = dependencyMapping.ExecuteContractReferences
+	} else if msgInfo.MessageType == acltypes.WasmMessageSubtype_QUERY && len(dependencyMapping.QueryContractReferences) != 0 {
+		specificContractRefs = dependencyMapping.QueryContractReferences
+	}
+	for _, specificContractRef := range specificContractRefs {
+		if specificContractRef.MessageName == msgInfo.MessageName {
+			contractRefs = append(contractRefs, specificContractRef.ContractReferences...)
```
codchen, line 189: break

x/accesscontrol/types/message_dependency_mapping.go
```diff
@@ -131,5 +133,51 @@ func ValidateWasmDependencyMapping(mapping acltypes.WasmDependencyMapping) error
 	if lastAccessOp.Operation.AccessType != acltypes.AccessType_COMMIT {
```
codchen, line 133: we can also verify there is no duplicate message name like we do with normal access ops

proto/cosmos/accesscontrol/accesscontrol.proto
```diff
@@ -18,6 +18,18 @@ message WasmAccessOperation {
     string selector = 3;
 }
 
+message WasmContractReference {
+    string contract_address = 1;
+    WasmMessageSubtype message_type = 2;
```
codchen, line 23: nit we can probably get rid of this field by having `contract_query_references` and `contract_execute_references` in `WasmContractReferences` to save a few bytes and some if statements later on in code

codchen (approved):

proto/cosmos/accesscontrol/accesscontrol.proto
udpatil: This is different, because the contract_query_references and contract_execute_references specify what branch to take for contract A based on the incoming WasmMessageInfo. This message type specifies the message type to send to the contract B GetWasmDependency and building that WasmMessageInfo

codchen: I'm thinking of each `WasmContractReferences` corresponds to one message type of contract A, and its fields `contract_query_references` would be a list of `<contract X, some message type of contract X>` tuple, and likewise for execute. Would that work?

udpatil: I'm thinking of each WasmContractReference as a pointer to the contract B or C or whatever and what message to send to that contract. Especially relevant when we add in the changes that allow for JSON translation, since if we have a list of contract X with message type of contract X, we would also need the JSON translation template too, so it make more sense to me to have WasmContractReference be containing all of the information for calling another contract

PR #135, opened 2023-01-13
[acl] Add JSON message translation for contract references

udpatil:
This allows contract references to specify translation templates for messages that they will send to other contracts. Using this template, we translate the existing message into the new message format, which should help reduce the resource scope computed for the contract reference.

Added unit tests for translator as well as keeper


x/accesscontrol/types/wasm_test.go
```diff
@@ -25,3 +25,31 @@ func TestExtractMessage(t *testing.T) {
...
+	translator := NewWasmMessageTranslator("senderAddress", "contractAddress", info)
+
+	// lets test a simple case where we want to make a new message to perform a "swap"
+	jsonTranslationTemplate := []byte("{\"swap\":[\"_sender\",\"_contract_address\",\"__sender\",\".send.from\",[\".send.amount\"],{\"some_field\": \"_sender\"}]}")
+
+	expectedJson := []byte("{\"swap\":[\"senderAddress\",\"contractAddress\",\"sender\",\"bobAddress\",[10],{\"some_field\":\"senderAddress\"}]}")
+	outputJson, err := translator.TranslateMessageBody(jsonTranslationTemplate)
+	require.NoError(t, err)
+	require.Equal(t, expectedJson, outputJson)
+}
```
codchen, line 55: can we add a test case that involves `null` in json? just want to make sure it would cause crash

x/accesscontrol/keeper/keeper.go
```diff
@@ -209,22 +209,28 @@ func (k Keeper) GetWasmDependencyAccessOps(ctx sdk.Context, contractAddress sdk.
...
 	importedAccessOps := types.NewEmptyAccessOperationSet()
 
+	jsonTranslator := types.NewWasmMessageTranslator(senderBech, contractAddr.String(), msgInfo)
 	for _, contractReference := range contractReferences {
 		// We use this to import the dependencies from another contract address
 		importContractAddress, err := sdk.AccAddressFromBech32(contractReference.ContractAddress)
 		if err != nil {
 			return nil, err
 		}
 		// TODO: build new msgbody for the new contract execute / query msg in later milestone tasks based on the translation pattern
```
codchen, line 219: rm

codchen (approved):

x/accesscontrol/types/wasm_test.go
udpatil: do you mean where the template has null or the original JSOn message has a `null` that's picked up by the translator's JQ selection ?

PR #136, opened 2023-01-18
Add eventManager to mergeIterator

LCyson:
This PR adds eventMangaer to the mergeIterator to record resource access for READ operation that doesn't access to the cache iterator but the parent (often iavl) iterator.

- Load test
- Tested on local node
- unit test

### DIFF
--- a/store/cachekv/mergeiterator.go
+++ b/store/cachekv/mergeiterator.go
@@ -137,7 +137,9 @@ func (iter *cacheMergeIterator) Value() []byte {
 
 	// If parent is invalid, get the cache value.
 	if !iter.parent.Valid() {
-		return iter.cache.Value()
+		value := iter.cache.Value()
+		iter.eventManager.EmitResourceAccessReadEvent("iterator", iter.storeKey, iter.cache.Key(), value)
+		return value
 	}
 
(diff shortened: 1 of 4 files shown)


store/cachekv/mergeiterator.go
```diff
@@ -132,7 +142,9 @@ func (iter *cacheMergeIterator) Value() []byte {
 
 	// If cache is invalid, get the parent value.
 	if !iter.cache.Valid() {
-		return iter.parent.Value()
+		value := iter.parent.Value()
+		iter.eventManager.EmitResourceAccessReadEvent("iterator", iter.storeKey, iter.parent.Key(), value)
```
codchen, line 146: it might be cleaner to just remove the emit logic in memiterator and always emit here in the mergeiterator

LCyson: make senses, was there a reason why we only added it to the memiterator at the first place?

BrandonWeng: I added it there, it was a misunderstanding. We should just always emit

BrandonWeng (approved):

codchen (approved):

PR #142, opened 2023-01-22
Add mutex to CommitKVStoreCache access

codchen:
Since the same CommitKVStoreCache can be accessed concurrently by multiple goroutines, we need to protect its access by a mutex. Not using a RWMutex here because even `Get` potentially involves an `Add` operation.

existing unit tests which are all synchronous should still hold
will test on clusters

### DIFF
--- a/store/cache/cache.go
+++ b/store/cache/cache.go
@@ -2,6 +2,7 @@ package cache
 
 import (
 	"fmt"
+	"sync"
 
 	"github.com/cosmos/cosmos-sdk/store/cachekv"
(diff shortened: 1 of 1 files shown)


BrandonWeng (approved):
I wonder why we didn't see this before... 2.0.0beta cluster ran for multiple weeks with no issues

codchen:
> I wonder why we didn't see this before... 2.0.0beta cluster ran for multiple weeks with no issues

I think this could only happen the intermediate steps of cache operation are compromised, which should be fairly rare unless traffic is high, but I'm not sure

PR #143, opened 2023-01-24
[acl] Allow parsing contract ref addrs from jq or sender

udpatil:
This allows for parsing the contract address from the message body / sender address for use cases such as adapter contracts that many contracts may call that fan out based on which contract called / if it sent a contract address to call

Added unit test for contract address parsing


x/accesscontrol/keeper/keeper.go
```diff
@@ -206,13 +207,40 @@ func (k Keeper) GetWasmDependencyAccessOps(ctx sdk.Context, contractAddress sdk.
...
+	if err != nil {
+		return maybeContractAddress
+	}
+	// if we parse it properly convert to string and return
+	var newValBytes string
+	err = json.Unmarshal(data, &newValBytes)
+	if err != nil {
+		return maybeContractAddress
+	}
+	return newValBytes
```
codchen, line 233: if this returns a nested json, the `AccAddressFromBech32` below would error. In that case do we want to propagate the error or use maybeContractAddress instead?

codchen (approved):

x/accesscontrol/keeper/keeper.go
udpatil: I think it would be correct to error, since that would indicate that the wrong JQ selector path was used and while it was a valid path it didn't build a proper contract address

PR #147, opened 2023-01-27
Fix application state rollback logic

codchen:
- Fix rollback logic so that states of up to `target` height, instead of `target - 1`, are kept, since we want application state height to be on the same height of tendermint state height (which equals `target`) post rollback
- Commit state without bumping version. Otherwise the application state will end up in `target + 1` (and of the exact same data as height `target`) 

tested with local sei

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -137,6 +137,8 @@ replace (
 	github.com/99designs/keyring => github.com/cosmos/keyring v1.1.7-0.20210622111912-ef00f8ac3d76
 	github.com/confio/ics23/go => github.com/cosmos/cosmos-sdk/ics23/go v0.8.0
 
+	github.com/cosmos/iavl => github.com/sei-protocol/sei-iavl v0.0.1
+
 	// Fix upstream GHSA-h395-qcrw-5vmq vulnerability.
 	// TODO Remove it: https://github.com/cosmos/cosmos-sdk/issues/10409
 	github.com/gin-gonic/gin => github.com/gin-gonic/gin v1.7.0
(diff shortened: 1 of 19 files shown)


store/iavl/store.go
```diff
@@ -128,10 +128,17 @@ func (st *Store) GetWorkingHash() ([]byte, error) {
 
 // Commit commits the current store state and returns a CommitID with the new
 // version and hash.
```
yzang2019, line 130: Can we add a comment here for when we should bump vs not bumping the version?

yzang2019 (approved):
LGTM

go.mod
```diff
@@ -136,6 +136,8 @@ replace (
 	github.com/99designs/keyring => github.com/cosmos/keyring v1.1.7-0.20210622111912-ef00f8ac3d76
 	github.com/confio/ics23/go => github.com/cosmos/cosmos-sdk/ics23/go v0.8.0
 
+	github.com/cosmos/iavl => github.com/sei-protocol/sei-iavl v0.0.1
```
philipsu522, line 139: nice

store/rootmulti/store_test.go
```diff
@@ -82,7 +82,7 @@ func TestCacheMultiStoreWithVersion(t *testing.T) {
 	store1 := ms.GetStoreByName("store1").(types.KVStore)
```
philipsu522, line 82: is it possible / easy to add a unit test for the case where we use `Commit(false)`? if not it's fine

philipsu522 (approved):

store/rootmulti/store_test.go
codchen: good point. Added one

PR #150, opened 2023-01-31
Reset interblock cache after statesync completes

codchen:
Interblock cache might be polluted with pre-statesync entries that are sticky after statesync completes. Resetting the cache to empty once statesync finishes would solve this issue.

tested on ec2-3-73-76-195.eu-central-1.compute.amazonaws.com with steps in https://linear.app/seilabs/issue/SEI-2800/%5Bdevnet-3-internal%5D-investigate-state-sync-error-xdistribution


yzang2019 (approved):
Nice finding!

philipsu522 (approved):

PR #154, opened 2023-02-03
[go.mod][lru] bump LRU to v2.0.1 for generics

BrandonWeng:
Bump golang LRU to use newer version with Generics 

We suspect that there's an unsafe conversion between the byte key in the map code. With the newer version, it supports generics so we need to define the types for the underlying LRU instead of interface (where we suspect the issue to be)

Ran two clusters, one 35 node one and a smaller one. Saw app hash issues but none were related to the LRU issue we saw  and the app hash was repro'd in a cluster not running this new version so it's unlikely to be releated. 

image

image

### DIFF
--- a/store/cache/cache.go
+++ b/store/cache/cache.go
@@ -28,7 +28,7 @@ type (
 	// CommitKVStore and below is completely irrelevant to this layer.
 	CommitKVStoreCache struct {
 		types.CommitKVStore
-		cache       *lru.TwoQueueCache
+		cache       *lru.TwoQueueCache[string, []byte]
 		cacheKVSize int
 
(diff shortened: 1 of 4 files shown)


store/cache/cache.go
```diff
@@ -28,7 +28,7 @@ type (
 	// CommitKVStore and below is completely irrelevant to this layer.
 	CommitKVStoreCache struct {
 		types.CommitKVStore
-		cache       *lru.TwoQueueCache
+		cache       *lru.TwoQueueCache[string, []byte]
```
BrandonWeng, line 31: []byte is not comparable

philipsu522 (approved):

PR #161, opened 2023-02-09
Update staking default params

BrandonWeng:
Lowering the threshold for which the power ratio is enforced to make sure we have a power cap in place.

Note this means we can't emergency delegate ourselves majority voting power in newer chains 

Tested on local chain 
image

### DIFF
--- a/types/staking.go
+++ b/types/staking.go
@@ -10,9 +10,8 @@ const (
 	// Max percentage of the total voting power a single validator can have
 	DefaultMaxVotingPowerRatio = "0.1"
 
-	// Maximum amount of staking power we allow per validator
-	// before the max voting power ratio is enforced
-	// Treshold for number of token staked = DefaultPowerReduction * DefaultMaxVotingPowerEnforcementThreshold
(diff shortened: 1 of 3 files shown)


types/staking.go
```diff
@@ -7,10 +7,13 @@ const (
...
 	// default maximal voting power ratio
+	// Max percentage of the total voting power a single validator can have
 	DefaultMaxVotingPowerRatio = "0.1"
 
-	// default voting power enforcement threshold
-	DefaultMaxVotingPowerEnforcementThreshold uint64 = 10000000000000
+	// Maximum amount of staking power we allow per validator
+	// before the max voting power ratio is enforced
+	// Treshold for number of token staked = DefaultPowerReduction * DefaultMaxVotingPowerEnforcementThreshold
+	DefaultMaxVotingPowerEnforcementThreshold uint64 = 100
```
philipsu522, line 16: lets do something like 10k for now, since we're having folks start off with 10

philipsu522 (approved):

LCyson (approved):

PR #166, opened 2023-02-15
Add flag for profiling

BrandonWeng:
```
> seid start --profile
9:08AM INF Listening for profiling at http://localhost:6060/debug/pprof/
9:08AM INF starting node with ABCI Tendermint in-process
9:08AM INF starting service impl=Node service=Node
9:08AM INF starting service impl=proxyClient module=proxy service=proxyClient
9:08AM INF starting service impl=localClient service=localClient
```

```
> seid start --profile=false
9:08AM INF starting node with ABCI Tendermint in-process
9:08AM INF starting service impl=Node service=Node
9:08AM INF starting service impl=proxyClient module=proxy service=proxyClient
9:08AM INF starting service impl=localClient service=localClient
9:08AM INF starting service impl=EventBus module=eventbus service=EventBus
```

### DIFF
--- a/server/start.go
+++ b/server/start.go
@@ -186,6 +186,7 @@ is performed. Note, when enabled, gRPC will also be automatically enabled.
 	cmd.Flags().Bool(FlagInterBlockCache, true, "Enable inter-block caching")
 	cmd.Flags().String(flagCPUProfile, "", "Enable CPU profiling and write to the provided file")
 	cmd.Flags().Bool(FlagTrace, false, "Provide full stack traces for errors in ABCI Log")
+	// TODO(bweng):: disable by default for mainnet
 	cmd.Flags().Bool(FlagProfile, true, "Enable Profiling in the application")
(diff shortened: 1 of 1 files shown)


server/start.go
```diff
@@ -171,6 +186,8 @@ is performed. Note, when enabled, gRPC will also be automatically enabled.
 	cmd.Flags().Bool(FlagInterBlockCache, true, "Enable inter-block caching")
 	cmd.Flags().String(flagCPUProfile, "", "Enable CPU profiling and write to the provided file")
 	cmd.Flags().Bool(FlagTrace, false, "Provide full stack traces for errors in ABCI Log")
+	// TODO(bweng):: disable by default for mainnet
+	cmd.Flags().Bool(FlagProfile, true, "Enable Profiling in the application")
```
yzang2019, line 190: I think here we can just set default to false, and in our sei-infra, we can modify the systemd service to turn it on by default. wdyt?

BrandonWeng: Yeah sounds good

yzang2019 (approved):

PR #167, opened 2023-02-18
[staking] max voting threshold fix

BrandonWeng:
* fix exploit where first delegation can be as high as they want
* fix calculation where we need to take the power reduction of the new bond amount into account or else it will always be much larger than the ratio so it'll always fail once we cross the threshold

Additional unit test and verified it's able to start chain locally

### DIFF
--- a/types/staking.go
+++ b/types/staking.go
@@ -8,7 +8,7 @@ const (
 
 	// default maximal voting power ratio
 	// Max percentage of the total voting power a single validator can have
-	DefaultMaxVotingPowerRatio = "0.1"
+	DefaultMaxVotingPowerRatio = "0.2"
 
 	// Maximum amount of staking power we allow per validator before the max voting power ratio is enforced
 	// Threshold for number of token staked = DefaultPowerReduction * DefaultMaxVotingPowerEnforcementThreshold
--- a/types/staking.go
+++ b/types/staking.go
@@ -12,7 +12,7 @@ const (
 
 	// Maximum amount of staking power we allow per validator before the max voting power ratio is enforced
 	// Threshold for number of token staked = DefaultPowerReduction * DefaultMaxVotingPowerEnforcementThreshold
-	DefaultMaxVotingPowerEnforcementThreshold uint64 = 10000
+	DefaultMaxVotingPowerEnforcementThreshold uint64 = 1000000
 
 	// Delay, in blocks, between when validator updates are returned to the
 	// consensus-engine and when they are applied. For example, if
--- a/x/staking/keeper/delegation.go
+++ b/x/staking/keeper/delegation.go
@@ -653,9 +653,8 @@ func (k Keeper) Delegate(
 		// Validator's new total power cannot exceed the max power ratio that's allowed
 		newVotingPowerRatio := validatorNewTotalPower.ToDec().Quo(newTotalPower.ToDec())
 		maxVotingPowerRatio := k.MaxVotingPowerRatio(ctx)
-		k.Logger(ctx).Debug("new total power ratio exceeds the max allowed ratio: %f > %f\n", newVotingPowerRatio, maxVotingPowerRatio)
 		if newVotingPowerRatio.GT(maxVotingPowerRatio) {
-			k.Logger(ctx).Error("validator's voting power ratio exceeds the max allowed ratio: %s > %s\n", newVotingPowerRatio, maxVotingPowerRatio)
+			k.Logger(ctx).Error("validator's voting power ratio exceeds the max allowed ratio: %s > %s\n", newVotingPowerRatio.TruncateInt().ToDec().String(), maxVotingPowerRatio.TruncateInt().ToDec().String())
 			return sdk.ZeroDec(), types.ErrExceedMaxVotingPowerRatio
 		}
 	}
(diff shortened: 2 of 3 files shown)


types/staking.go
```diff
@@ -8,11 +8,11 @@ const (
 
 	// default maximal voting power ratio
 	// Max percentage of the total voting power a single validator can have
-	DefaultMaxVotingPowerRatio = "0.1"
+	DefaultMaxVotingPowerRatio = "0.2"
 
 	// Maximum amount of staking power we allow per validator before the max voting power ratio is enforced
 	// Threshold for number of token staked = DefaultPowerReduction * DefaultMaxVotingPowerEnforcementThreshold
-	DefaultMaxVotingPowerEnforcementThreshold uint64 = 10000
+	DefaultMaxVotingPowerEnforcementThreshold uint64 = 1000000
```
BrandonWeng, line 15: Allows for more flexibility for our chains, 10k is the default we stake in most of our scripts for external validators

types/staking.go
```diff
@@ -8,11 +8,11 @@ const (
 
 	// default maximal voting power ratio
 	// Max percentage of the total voting power a single validator can have
-	DefaultMaxVotingPowerRatio = "0.1"
+	DefaultMaxVotingPowerRatio = "0.2"
```
BrandonWeng, line 11: 10% is hit pretty easily when there's little validators, set to 20% and allow the param to be tuned in the future

x/staking/keeper/delegation.go
```diff
@@ -643,9 +643,22 @@ func (k Keeper) Delegate(
...
-		if newVotingPowerRatio.GT(k.MaxVotingPowerRatio(ctx)) {
+
+	// 1 power = Bond Amount / Power Reduction
+	validatorAddtionalPower := bondAmt.Quo(k.PowerReduction(ctx))
+	newTotalPower := lastTotalPower.Add(validatorAddtionalPower)
+
+	// If it's beyond genesis then enforce power ratio per validator if there's more than maxVotingPowerEnforcementThreshold
+	if newTotalPower.GTE(maxVotingPowerEnforcementThreshold) && ctx.BlockHeight() > 0 {
+		// Convert bond amount to power first
+		validatorNewTotalPower := validator.Tokens.Add(bondAmt).Quo(k.PowerReduction(ctx))
```
BrandonWeng, line 654: this was the issue before, we didn't convert bond amt to power

x/staking/keeper/delegation.go
```diff
@@ -643,9 +643,22 @@ func (k Keeper) Delegate(
...
-	if lastTotalPower.GTE(maxVotingPowerEnforcementThreshold) {
-		newVotingPowerRatio := sdk.NewDecFromInt(validator.Tokens.Add(bondAmt)).Quo(sdk.NewDecFromInt(lastTotalPower))
-		if newVotingPowerRatio.GT(k.MaxVotingPowerRatio(ctx)) {
+
+	// 1 power = Bond Amount / Power Reduction
+	validatorAddtionalPower := bondAmt.Quo(k.PowerReduction(ctx))
+	newTotalPower := lastTotalPower.Add(validatorAddtionalPower)
+
+	// If it's beyond genesis then enforce power ratio per validator if there's more than maxVotingPowerEnforcementThreshold
+	if newTotalPower.GTE(maxVotingPowerEnforcementThreshold) && ctx.BlockHeight() > 0 {
```
BrandonWeng, line 652: Also check that the new power will be more than the enforcement or else the first delegation can be arbitrarily large

BrandonWeng: Also don't check if this is genesis block as gentx will fail since delegate is called per validator sequentially , and the first couple will have more power ratio naturally

x/staking/keeper/delegation.go
```diff
@@ -643,9 +643,22 @@ func (k Keeper) Delegate(
...
+		// Convert bond amount to power first
+		validatorNewTotalPower := validator.Tokens.Add(bondAmt).Quo(k.PowerReduction(ctx))
+		// Validator's new total power cannot exceed the max power ratio that's allowed
+		newVotingPowerRatio := validatorNewTotalPower.ToDec().Quo(newTotalPower.ToDec())
```
udpatil, line 656: how is this expected to behave if for example I send a TX with multiple messages that delegates to ALL validators (one by one in each message) such that the intended end of transaction is for them to still be even, but for example after the first message is executed the voting power would be imbalanced? The transaction would fail, right? is that intended behavior?

x/staking/keeper/delegation.go
udpatil, line 652: is the intended behavior here that voting power caps will only be enforced AFTER the max voting power exceeds the threshold, and prior to that it can be as imbalanced as it wants to be?

udpatil (approved):

x/staking/keeper/delegation.go
BrandonWeng: Based on the original design that does seem to be the case, the idea is that the power ratio will only be enforced after the chain exceeds a certain total power threshold. 

cc @LCyson

x/staking/keeper/delegation.go
BrandonWeng: Yeah, the whole TX would fail. 

If a message is invalid then the TX itself is invalid and should fail right? If TXs aren't atomic then it makes the state transition non-deterministic for the user

philipsu522 (approved):

x/staking/keeper/delegation.go
LCyson: this was for allowing us to have flexibility to decide when we want to enforce this voting max cap (esp for the first few blocks if we want to delegate majority voting power to our own nodes for example). But for mainnet the best practice is prob to start enforcing the max power after the genesis block

LCyson (approved):

PR #179, opened 2023-03-09
fix params race condition

codchen:
Detected by Go race detector

### DIFF
--- a/types/events.go
+++ b/types/events.go
@@ -7,6 +7,7 @@ import (
 	"reflect"
 	"sort"
 	"strings"
+	"sync"
 
 	"golang.org/x/exp/maps"
 	"golang.org/x/exp/slices"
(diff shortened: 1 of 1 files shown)


x/params/types/subspace.go
```diff
@@ -21,23 +21,26 @@ const (
...
-		key:         key,
-		tkey:        tkey,
-		name:        []byte(name),
-		table:       NewKeyTable(),
+		cdc:          cdc,
+		legacyAmino:  legacyAmino,
+		key:          key,
+		tkey:         tkey,
+		name:         nameBz,
+		suffixedName: append(nameBz, '/'),
```
yzang2019, line 42: why do we add a suffixed name here?

philipsu522 (approved):

PR #180, opened 2023-03-09
Reset Eventmanager upon runTx 

BrandonWeng:
Overwrite the event manager after each runTx since the events are only used in each runTx for parallelTx validation 

Unit tests

### DIFF
--- a/baseapp/test_helpers.go
+++ b/baseapp/test_helpers.go
@@ -17,6 +17,9 @@ func (app *BaseApp) Check(txEncoder sdk.TxEncoder, tx sdk.Tx) (sdk.GasInfo, *sdk
 	}
 	ctx := app.checkState.ctx.WithTxBytes(bz).WithVoteInfos(app.voteInfos).WithConsensusParams(app.GetConsensusParams(app.checkState.ctx))
 	gasInfo, result, _, _, err := app.runTx(ctx, runTxModeCheck, bz)
+	if len(ctx.MultiStore().GetEvents()) > 0 {
+		panic("Expected checkTx events to be empty")
+	}
 	return gasInfo, result, err
 }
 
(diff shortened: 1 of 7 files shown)


baseapp/test_helpers.go
```diff
@@ -35,6 +41,10 @@ func (app *BaseApp) Deliver(txEncoder sdk.TxEncoder, tx sdk.Tx) (sdk.GasInfo, *s
 	}
 	ctx := app.deliverState.ctx.WithTxBytes(bz).WithVoteInfos(app.voteInfos).WithConsensusParams(app.GetConsensusParams(app.deliverState.ctx))
 	gasInfo, result, _, _, err := app.runTx(ctx, runTxModeDeliver, bz)
+
+	if len(ctx.MultiStore().GetEvents()) > 0 {
+		panic("Expected deliverTx events to be empty")
+	}
```
BrandonWeng, line 47: Before the reset change this would fail

baseapp/test_helpers.go
```diff
@@ -17,13 +17,19 @@ func (app *BaseApp) Check(txEncoder sdk.TxEncoder, tx sdk.Tx) (sdk.GasInfo, *sdk
 	}
 	ctx := app.checkState.ctx.WithTxBytes(bz).WithVoteInfos(app.voteInfos).WithConsensusParams(app.GetConsensusParams(app.checkState.ctx))
 	gasInfo, result, _, _, err := app.runTx(ctx, runTxModeCheck, bz)
+	if len(ctx.MultiStore().GetEvents()) > 0 {
+		panic("Expected checkTx events to be empty")
```
philipsu522, line 21: does it make sense to panic if its just events? this isn't on the critical path so it seems strange to error out b/c of this. Perhaps a warning log instead?

BrandonWeng: yeah, this is just unit test

yzang2019 (approved):
Verified in a previously OOM node and confirmed the issue got fixed, thanks!

baseapp/test_helpers.go
philipsu522: oo ops

philipsu522 (approved):

PR #186, opened 2023-03-10
Reset Events

BrandonWeng:
See inline comments

My thought process was that the ctx passed to Enblock is shared in all of ProcessBlock. So any store in the begin/mid block would be branching from that ctx's store and cache misses in the child stores would go to the parent store and store a read event.

```
=
==================
WARNING: DATA RACE
Read at 0x00c00b1146f8 by goroutine 5537:
  github.com/cosmos/cosmos-sdk/store/cachekv.(*Store).Set()
      /root/go/pkg/mod/github.com/sei-protocol/sei-cosmos@v0.2.1-race-cond/store/cachekv/store.go:133 +0x138
  github.com/cosmos/cosmos-sdk/store/cachekv.(*Store).Write()

Previous write at 0x00c00b1146f8 by goroutine 5431:
  github.com/cosmos/cosmos-sdk/store/cachekv.(*Store).ResetEvents()
      /root/go/pkg/mod/github.com/sei-protocol/sei-cosmos@v0.2.1-race-cond/store/cachekv/store.go:97 +0xd0

Goroutine 5537 (running) created at:
  github.com/sei-protocol/sei-chain/app.(*App).ProcessBlockConcurrent()
      /home/ubuntu/sei-chain/app/app.go:1107 +0x1fd

Goroutine 5431 (running) created at:
  github.com/sei-protocol/sei-chain/app.(*App).ProcessBlockConcurrent()
      /home/ubuntu/sei-chain/app/app.go:1107 +0x1fd
=================


```

Unit tests, and going deploy on a node that's stuck to verify OOOM issue

### DIFF
--- a/baseapp/abci.go
+++ b/baseapp/abci.go
@@ -179,6 +179,9 @@ func (app *BaseApp) MidBlock(ctx sdk.Context, height int64) (events []abci.Event
 
 // EndBlock implements the ABCI interface.
 func (app *BaseApp) EndBlock(ctx sdk.Context, req abci.RequestEndBlock) (res abci.ResponseEndBlock) {
+	// Clear DeliverTx Events
+	ctx.MultiStore().ResetEvents()
+
 	defer telemetry.MeasureSince(time.Now(), "abci", "end_block")
 
 	if app.endBlocker != nil {
--- a/baseapp/baseapp.go
+++ b/baseapp/baseapp.go
@@ -711,8 +711,11 @@ func (app *BaseApp) cacheTxContext(ctx sdk.Context, txBytes []byte) (sdk.Context
 // returned if the tx does not run out of gas and if all the messages are valid
 // and execute successfully. An error is returned otherwise.
 func (app *BaseApp) runTx(ctx sdk.Context, mode runTxMode, txBytes []byte) (gInfo sdk.GasInfo, result *sdk.Result, anteEvents []abci.Event, priority int64, err error) {
-	// Reset events after each runTx as its not used after the tx is processed
-	defer ctx.MultiStore().ResetEvents()
+	// Reset events after each checkTx or simulateTx or recheckTx
+	// DeliverTx is garbage collected after FinalizeBlocker
+	if mode != runTxModeDeliver {
+		defer ctx.MultiStore().ResetEvents()
+	}
 
 	// Wait for signals to complete before starting the transaction. This is needed before any of the
 	// resources are acceessed by the ante handlers and message handlers.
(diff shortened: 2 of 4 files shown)


baseapp/baseapp.go
```diff
@@ -711,8 +711,11 @@ func (app *BaseApp) cacheTxContext(ctx sdk.Context, txBytes []byte) (sdk.Context
 // returned if the tx does not run out of gas and if all the messages are valid
 // and execute successfully. An error is returned otherwise.
 func (app *BaseApp) runTx(ctx sdk.Context, mode runTxMode, txBytes []byte) (gInfo sdk.GasInfo, result *sdk.Result, anteEvents []abci.Event, priority int64, err error) {
-	// Reset events after each runTx as its not used after the tx is processed
-	defer ctx.MultiStore().ResetEvents()
+	// Reset events after each checkTx or simulateTx or recheckTx
+	// DeliverTx is garbage collected after FinalizeBlocker
+	if mode != runTxModeDeliver {
+		defer ctx.MultiStore().ResetEvents()
+	}
```
BrandonWeng, line 718: So the issue was that the deliverTx calls shared the same parent store, and MultiStore.ResetEvents() calls it on the parent store. 

So therefore when goroutine A is clearing events, it can also impact any other goroutine trying to emit events on the child store since its branched later

baseapp/abci.go
```diff
@@ -179,6 +179,9 @@ func (app *BaseApp) MidBlock(ctx sdk.Context, height int64) (events []abci.Event
 
 // EndBlock implements the ABCI interface.
 func (app *BaseApp) EndBlock(ctx sdk.Context, req abci.RequestEndBlock) (res abci.ResponseEndBlock) {
+	// Clear DeliverTx Events
+	ctx.MultiStore().ResetEvents()
```
BrandonWeng, line 183: Clear it in endBlock when we're done with the deliverTx events

baseapp/baseapp.go
BrandonWeng: The events aren't needed for simulate, checktx and rechecktx so we can clear it immediate after each runTx

yzang2019 (approved):

baseapp/abci.go
BrandonWeng: Maybe we don't really need to clear the events here and just let it get garbage collected once the block is committed since it's not interblock store but mainly reset for checktx (runs synchronously) so we dont OOM

philipsu522 (approved):

PR #187, opened 2023-03-10
Fix race condition in ABCI calls

codchen:
Move operations that mutate `app` fields into setter functions and protect by RWLock

WIP


baseapp/abci.go
```diff
@@ -678,9 +675,12 @@ func (app *BaseApp) createQueryContext(height int64, prove bool) (sdk.Context, e
 			)
 	}
 
+	// app.abciMtx.RLock()
```
philipsu522, line 678: is this still wip?}

philipsu522 (approved):

PR #191, opened 2023-03-14
Uday slashing parallel

udpatil:
Makes slashing begin blocker parallelized and also modify cachekv to acquire readlock instead of lock for GET and ITERATOR

DO NOT MERGE YET - still testing


store/cache/cache.go
```diff
@@ -106,22 +106,24 @@ func (ckv *CommitKVStoreCache) CacheWrap(storeKey types.StoreKey) types.CacheWra
...
 
 	types.AssertValidKey(key)
 
 	keyStr := string(key)
 	value, ok := ckv.cache.Get(keyStr)
 	if ok {
 		// cache hit
+		ckv.mtx.RUnlock()
 		return value
 	}
```
yzang2019, line 120: We need to unlock the readlock before acquire write lock here right?

udpatil: yes good call

yzang2019 (approved):

PR #195, opened 2023-03-15
Revert EventAttribute field types

LCyson:
This PR updates EventAttribute `key` and `value` data type. Please find details in this PR https://github.com/sei-protocol/sei-tendermint/pull/85

- unit test

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -145,7 +145,7 @@ replace (
 	github.com/gin-gonic/gin => github.com/gin-gonic/gin v1.7.0
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
 
-	github.com/tendermint/tendermint => github.com/sei-protocol/sei-tendermint v0.1.171
+	github.com/tendermint/tendermint => github.com/sei-protocol/sei-tendermint v0.1.179-0.20230315064114-c903df251edf
 
 	// latest grpc doesn't work with with our modified proto compiler, so we need to enforce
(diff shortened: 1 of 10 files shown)


go.mod
```diff
@@ -145,7 +145,7 @@ replace (
 	github.com/gin-gonic/gin => github.com/gin-gonic/gin v1.7.0
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
 
-	github.com/tendermint/tendermint => github.com/sei-protocol/sei-tendermint v0.1.171
+	github.com/tendermint/tendermint => github.com/sei-protocol/sei-tendermint v0.1.179-0.20230315064114-c903df251edf
```
BrandonWeng, line 148: Was this intentional?

BrandonWeng: Can we tag a new tendermint release instead

BrandonWeng (commented):
Should be fine for parallelTx validation if it's just changing this. 

Could you also spin up a loadtest cluster just to make sure that it doesn't have any untended side effects before we deploy to sei-devnet-3?

go.mod
philipsu522: +1. Let's revert this and the protocol-team can cut a new release

philipsu522 (approved):

go.mod
LCyson: yes will update it to a tm release after this is merged https://github.com/sei-protocol/sei-tendermint/pull/85

PR #198, opened 2023-03-16
Uday slashing refactor

udpatil:
This modifies the validator bit array to use one key that stores all validator bits instead of separate keys for each index

A couple things to look at closely: 
- The resizing logic in abci.go. 
  - This introduces the logic to resize the missed block array in case the signing window size changes
- migration
  - This adds logic to migrate the store from individual nodes to the new format.

Unit tests + migration unit tests


x/slashing/abci.go
```diff
@@ -56,6 +56,55 @@ func BeginBlocker(ctx sdk.Context, req abci.RequestBeginBlock, k keeper.Keeper)
 		if writeInfo == nil {
 			panic("Expected slashing write info to be non-nil")
 		}
+		// Check if we need to resize the array if there was recently a change in slashing window size
+		window := k.SignedBlocksWindow(ctx)
+		missedInfo, found := k.GetValidatorMissedBlocks(ctx, writeInfo.ConsAddr)
+		missedBlockLen := int64(len(missedInfo.MissedBlocks))
+		if found && window != missedBlockLen {
```
philipsu522, line 63: what does it mean if found != true?

x/slashing/abci.go
```diff
@@ -56,6 +56,55 @@ func BeginBlocker(ctx sdk.Context, req abci.RequestBeginBlock, k keeper.Keeper)
...
+			// we need to resize the missed block array AND update the signing info accordingly
+			switch {
+			case missedBlockLen < window:
+				// missed block array too short, lets expand it
+				indexOffset := writeInfo.SigningInfo.IndexOffset % window
```
philipsu522, line 68: isn't `writeInfo.SigningInfo.IndexOffset` always less than `window`?
If we're expanding from 5 -> 7, then wouldn't we just map everything from:
[0...5]
[0...5, nil, nil]
?

proto/cosmos/slashing/v1beta1/slashing.proto
```diff
@@ -32,6 +32,13 @@ message ValidatorSigningInfo {
   int64 missed_blocks_counter = 6 [(gogoproto.moretags) = "yaml:\"missed_blocks_counter\""];
 }
 
+// Stores a sliding window of the last `signed_blocks_window` blocks indicating whether the validator missed the block
+message ValidatorMissedBlockArray {
+  string address = 1;
+  // Array of length signing signed blocks window that contains whether the validator missed the block at that index
```
philipsu522, line 38: how come we decided to store array of bool rather than just missed block heights?

x/slashing/abci.go
```diff
@@ -56,6 +56,55 @@ func BeginBlocker(ctx sdk.Context, req abci.RequestBeginBlock, k keeper.Keeper)
...
 		}
+		// Check if we need to resize the array if there was recently a change in slashing window size
+		window := k.SignedBlocksWindow(ctx)
+		missedInfo, found := k.GetValidatorMissedBlocks(ctx, writeInfo.ConsAddr)
+		missedBlockLen := int64(len(missedInfo.MissedBlocks))
+		if found && window != missedBlockLen {
+			// we need to resize the missed block array AND update the signing info accordingly
+			switch {
+			case missedBlockLen < window:
+				// missed block array too short, lets expand it
```
philipsu522, line 67: is the logic here, that if we're going from 5 -> 7 window:
- before: [0..5]
- after: [nil, nil, 0..5]
?

x/slashing/abci.go
```diff
@@ -56,6 +56,55 @@ func BeginBlocker(ctx sdk.Context, req abci.RequestBeginBlock, k keeper.Keeper)
...
+				// traverse backwards through the old array (smaller), and copy it to the relevant index offsets based on the new window and updated relative index offset with the new window size
+				for i := relativeIndexOffset + missedBlockLen; i > relativeIndexOffset; i-- {
+					missedBlockIdx := i % missedBlockLen
+					newMissedBlocks[indexOffsetCounter%window] = missedInfo.MissedBlocks[missedBlockIdx]
+					indexOffsetCounter--
+				}
+				missedInfo.MissedBlocks = newMissedBlocks
+				k.SetValidatorMissedBlocks(ctx, writeInfo.ConsAddr, missedInfo)
+			case missedBlockLen > window:
+				// missed block array too long, we need to trim
```
philipsu522, line 81: how come we dont' just take the most recent values?
i.e. if we resize 5 -> 3
- before: [0..5]
- after: [3..5]
?

proto/cosmos/slashing/v1beta1/slashing.proto
yzang2019: +1, this should be a slice of blockId right?

x/slashing/abci.go
```diff
@@ -56,6 +56,55 @@ func BeginBlocker(ctx sdk.Context, req abci.RequestBeginBlock, k keeper.Keeper)
...
+		missedInfo, found := k.GetValidatorMissedBlocks(ctx, writeInfo.ConsAddr)
+		missedBlockLen := int64(len(missedInfo.MissedBlocks))
+		if found && window != missedBlockLen {
+			// we need to resize the missed block array AND update the signing info accordingly
+			switch {
+			case missedBlockLen < window:
+				// missed block array too short, lets expand it
+				indexOffset := writeInfo.SigningInfo.IndexOffset % window
+				relativeIndexOffset := writeInfo.SigningInfo.IndexOffset % missedBlockLen
+				newMissedBlocks := make([]bool, window)
```
yzang2019, line 70: Can we update the logic to use blockId?

x/slashing/abci.go
udpatil: no, index offset is always increasing, but then its used by modding with window size to get the new index I believe

x/slashing/abci.go
udpatil: the logic is such that our index offset should still point to the same value so that the next value to get overwritten is the oldest value

x/slashing/abci.go
udpatil: we do, but we need to make sure theyre ordered properly in the resulting array such that the validators index offset will still point to the correct value AFTER the mod with the new window size

x/slashing/abci.go
udpatil: yup, will do

x/slashing/keeper/infractions.go
```diff
@@ -34,29 +34,37 @@ func (k Keeper) HandleValidatorSignatureConcurrent(ctx sdk.Context, addr cryptot
...
-	switch {
-	case !previous && missed:
-		// Array value has changed from not missed to missed, increment counter
-		signInfo.MissedBlocksCounter++
-	case previous && !missed:
-		// Array value has changed from missed to not missed, decrement counter
-		signInfo.MissedBlocksCounter--
-	default:
-		// Array value at this index has not changed, no need to update counter
+	// previous = k.GetValidatorMissedBlockBitArray(ctx, consAddr, index)
```
philipsu522, line 42: let's update this comment + remove commented out code

philipsu522 (approved):
change looks good to me, especially with the unit tests. nice work

x/slashing/keeper/infractions.go
```diff
@@ -80,6 +85,7 @@ func (k Keeper) HandleValidatorSignatureConcurrent(ctx sdk.Context, addr cryptot
 	maxMissed := k.SignedBlocksWindow(ctx) - minSignedPerWindow
 	shouldSlash = false
 	// if we are past the minimum height and the validator has missed too many blocks, punish them
+	// TODO: this may be problematic if the signed blocks window was just reduced -> then missed blocks counter might be too high due to historical misses
```
yzang2019, line 88: If the window size changed, then when we do evictions, we will not count the old historical blocks, so this should not be a concern right?

proto/cosmos/slashing/v1beta1/slashing.proto
```diff
@@ -14,6 +14,23 @@ message ValidatorSigningInfo {
   option (gogoproto.equal)            = true;
   option (gogoproto.goproto_stringer) = false;
 
+  string address = 1;
+  // Height at which validator was first a candidate OR was unjailed
+  int64 start_height = 2 [(gogoproto.moretags) = "yaml:\"start_height\""];
```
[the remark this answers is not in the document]
udpatil: I believe this is used to provide a grace period in which the validator doesn't get jailed:
https://github.com/sei-protocol/sei-cosmos/blob/3983c683f8f27dc62c309aa0d5afa750d2e85010/x/slashing/keeper/infractions.go#L84

x/slashing/keeper/infractions.go
```diff
@@ -80,6 +85,7 @@ func (k Keeper) HandleValidatorSignatureConcurrent(ctx sdk.Context, addr cryptot
...
 	if height > minHeight && signInfo.MissedBlocksCounter > maxMissed {
```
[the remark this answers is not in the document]
udpatil: we still need height > minHeight since it provides a grace period in which a validator doesn't get jailed

x/slashing/keeper/infractions.go
udpatil: correct, removed the TODO,  i had initially put it there since i wasnt certain based on the old prior approach

yzang2019 (approved):

PR #200, opened 2023-03-18
Emit transaction and message metrics in endblock

BrandonWeng:
Issue with the other metrics is that it's being emitted during process block and checktx calls so the TX calls is inflated and not entirely representative of the TXs/Msgs/Orders processed per block

We branch the ctx in each, similar to the IAVL store logic, we only emit a single set of values, in this case whatever reaches endblock

Deferring the metrics to be emitted in sei-chain as we cannot guarantee the commit here 
 
Pretty similar to what the seid process is logging out, note that metrics will differ per instance depending on the scrape interval 

image

image


baseapp/baseapp.go
```diff
@@ -793,6 +793,8 @@ func (app *BaseApp) cacheTxContext(ctx sdk.Context, txBytes []byte) (sdk.Context
 // returned if the tx does not run out of gas and if all the messages are valid
 // and execute successfully. An error is returned otherwise.
 func (app *BaseApp) runTx(ctx sdk.Context, mode runTxMode, txBytes []byte) (gInfo sdk.GasInfo, result *sdk.Result, anteEvents []abci.Event, priority int64, err error) {
+	defer ctx.ContextMemCache().IncrMetricCounter(1, sdk.TX_COUNT)
```
BrandonWeng, line 796: include failed TXs and Msgs

yzang2019 (approved):

philipsu522 (approved):

PR #203, opened 2023-03-21
Hot fix for max gas change (#202)

BrandonWeng:
parallel TX with max gas causes TX to fail nondeterministically once the limit is exceeded because they're being executed concurrently

This is a hotfix that should be rolled back


udpatil (changes requested):
Looks good to me, changing the approval to request changes so it doesnt get accidentally merged

PR #207, opened 2023-03-27
[slashing] Refactor slashing to use bit packing

udpatil:
This refactors the slashing missed block tracking to use bit packing.

Mostly tested, still writing some unit tests

### DIFF
--- a/proto/cosmos/slashing/v1beta1/genesis.proto
+++ b/proto/cosmos/slashing/v1beta1/genesis.proto
@@ -23,13 +23,29 @@ message GenesisState {
 }
 
 // GenesisState defines the slashing module's genesis state.
+message GenesisStateLegacyMissingHeights {
+  // params defines all the paramaters of related to deposit.
+  Params params = 1 [(gogoproto.nullable) = false];
+
+  // signing_infos represents a map between validator addresses and their
+  // signing infos.
+  repeated SigningInfo signing_infos = 2
+      [(gogoproto.moretags) = "yaml:\"signing_infos\"", (gogoproto.nullable) = false];
+
+  // missed_blocks represents a map between validator addresses and their
+  // missed blocks.
+  repeated ValidatorMissedBlockArrayLegacyMissedHeights missed_blocks = 3
+      [(gogoproto.moretags) = "yaml:\"missed_blocks\"", (gogoproto.nullable) = false];
+}
+
+// GenesisState defines the slashing module's genesis state.
 message GenesisStateLegacyV43 {
   // params defines all the paramaters of related to deposit.
   Params params = 1 [(gogoproto.nullable) = false];
 
   // signing_infos represents a map between validator addresses and their
   // signing infos.
-  repeated SigningInfoLegacyV43 signing_infos = 2
+  repeated SigningInfo signing_infos = 2
       [(gogoproto.moretags) = "yaml:\"signing_infos\"", (gogoproto.nullable) = false];
 
   // missed_blocks represents a map between validator addresses and their
@@ -39,21 +55,21 @@ message GenesisStateLegacyV43 {
 }
 
 // SigningInfo stores validator signing info of corresponding address.
-message SigningInfoLegacyV43 {
-  // address is the validator address.
-  string address = 1;
-  // validator_signing_info represents the signing info of this validator.
-  ValidatorSigningInfoLegacyV43 validator_signing_info = 2
-      [(gogoproto.nullable) = false, (gogoproto.moretags) = "yaml:\"validator_signing_info\""];
-}
-
-
-// SigningInfo stores validator signing info of corresponding address.
 message SigningInfo {
   // address is the validator address.
   string address = 1;
   // validator_signing_info represents the signing info of this validator.
   ValidatorSigningInfo validator_signing_info = 2
+      [(gogoproto.nullable) = false, (gogoproto.moretags) = "yaml:\"validator_signing_info\""];
+}
+
+
+// SigningInfo stores validator signing info of corresponding address.
+message SigningInfoLegacyMissedHeights {
+  // address is the validator address.
+  string address = 1;
+  // validator_signing_info represents the signing info of this validator.
+  ValidatorSigningInfoLegacyMissedHeights validator_signing_info = 2
       [(gogoproto.nullable) = false, (gogoproto.moretags) = "yaml:\"validator_signing_info\""];
 }
 
--- a/proto/cosmos/slashing/v1beta1/genesis.proto
+++ b/proto/cosmos/slashing/v1beta1/genesis.proto
@@ -8,7 +8,7 @@ import "cosmos/slashing/v1beta1/slashing.proto";
 
 // GenesisState defines the slashing module's genesis state.
 message GenesisState {
-  // params defines all the paramaters of related to deposit.
+  // params defines all the paramaters of related to slashing.
   Params params = 1 [(gogoproto.nullable) = false];
 
   // signing_infos represents a map between validator addresses and their
@@ -24,7 +24,7 @@ message GenesisState {
 
 // GenesisState defines the slashing module's genesis state.
 message GenesisStateLegacyMissingHeights {
-  // params defines all the paramaters of related to deposit.
+  // params defines all the paramaters of related to slashing.
   Params params = 1 [(gogoproto.nullable) = false];
 
   // signing_infos represents a map between validator addresses and their
@@ -40,7 +40,7 @@ message GenesisStateLegacyMissingHeights {
 
 // GenesisState defines the slashing module's genesis state.
 message GenesisStateLegacyV43 {
-  // params defines all the paramaters of related to deposit.
+  // params defines all the paramaters of related to slashing.
   Params params = 1 [(gogoproto.nullable) = false];
 
   // signing_infos represents a map between validator addresses and their
--- a/proto/cosmos/slashing/v1beta1/slashing.proto
+++ b/proto/cosmos/slashing/v1beta1/slashing.proto
@@ -8,11 +8,8 @@ import "gogoproto/gogo.proto";
 import "google/protobuf/duration.proto";
 import "google/protobuf/timestamp.proto";
 
-// ValidatorSigningInfo defines a validator's signing info for monitoring their
-// liveness activity.
-message ValidatorSigningInfo {
+message ValidatorSigningInfoLegacyMissedHeights {
   option (gogoproto.equal)            = true;
-  option (gogoproto.goproto_stringer) = false;
 
   string address = 1;
   // Height at which validator was first a candidate OR was unjailed
@@ -28,8 +25,11 @@ message ValidatorSigningInfo {
   int64 missed_blocks_counter = 5 [(gogoproto.moretags) = "yaml:\"missed_blocks_counter\""];
 }
 
-message ValidatorSigningInfoLegacyV43 {
+// ValidatorSigningInfo defines a validator's signing info for monitoring their
+// liveness activity.
+message ValidatorSigningInfo {
   option (gogoproto.equal)            = true;
+  option (gogoproto.goproto_stringer) = false;
 
   string address = 1;
   // Height at which validator was first a candidate OR was unjailed
@@ -50,10 +50,19 @@ message ValidatorSigningInfoLegacyV43 {
 }
 
 // Stores a sliding window of the last `signed_blocks_window` blocks indicating whether the validator missed the block
-message ValidatorMissedBlockArray {
+message ValidatorMissedBlockArrayLegacyMissedHeights {
   string address = 1;
   // Array of contains the heights when the validator missed the block
   repeated int64 missed_heights = 2 [(gogoproto.moretags) = "yaml:\"missed_heights\""];
+}
+
+// Stores a sliding window of the last `signed_blocks_window` blocks indicating whether the validator missed the block
+message ValidatorMissedBlockArray {
+  string address = 1;
+  // store this in case window size changes but doesn't affect number of bit groups
+  int64 window_size = 2;
+  // Array of contains the missed block bits packed into uint64s
+  repeated uint64 missed_blocks = 3 [(gogoproto.moretags) = "yaml:\"missed_blocks\""];
 }
 
 // Params represents the parameters used for by the slashing module.
--- a/x/slashing/keeper/infractions.go
+++ b/x/slashing/keeper/infractions.go
@@ -36,30 +36,105 @@ func (k Keeper) HandleValidatorSignatureConcurrent(ctx sdk.Context, addr cryptot
 
 	window := k.SignedBlocksWindow(ctx)
 
+	index := signInfo.IndexOffset % k.SignedBlocksWindow(ctx)
+
 	missedInfo, found = k.GetValidatorMissedBlocks(ctx, consAddr)
+
 	if !found {
+		arrLen := window / 64
+		if window%64 > 0 {
+			arrLen += 1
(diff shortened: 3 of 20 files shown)


x/slashing/keeper/infractions.go
```diff
@@ -36,30 +36,105 @@ func (k Keeper) HandleValidatorSignatureConcurrent(ctx sdk.Context, addr cryptot
...
-			numberEvicted += 1
-		} else {
-			break
+	// TODO:  resize if necessary?
+	if found && missedInfo.WindowSize != window {
+		// we need to resize the missed block array AND update the signing info accordingly
+		switch {
+		case missedInfo.WindowSize < window:
+			// missed block array too short, lets expand it
+			boolArray := k.ParseBitGroupsToBoolArray(missedInfo.MissedBlocks, missedInfo.WindowSize)
```
yzang2019, line 60: Why do we parse bits into boolean array instead of just appending more uint64 to the bitsArray? And I think the below logics to copy things are also not needed when expanding the window.

x/slashing/keeper/infractions.go
```diff
@@ -36,30 +36,105 @@ func (k Keeper) HandleValidatorSignatureConcurrent(ctx sdk.Context, addr cryptot
...
+			oldIndex := signInfo.IndexOffset % missedInfo.WindowSize
```
yzang2019, line 61: Let's repurpose the index to be a circular index, it should never go beyond the window size limit. Then Index shouldn't need to change at all if we are expanding the window right?

x/slashing/keeper/infractions.go
```diff
@@ -36,30 +36,105 @@ func (k Keeper) HandleValidatorSignatureConcurrent(ctx sdk.Context, addr cryptot
...
+			finalBoolArr := make([]bool, window)
+			copy(finalBoolArr, newArray[window-shiftDiff:])
+			copy(finalBoolArr[shiftDiff:], newArray[:window-shiftDiff])
+			missedInfo.MissedBlocks = k.ParseBoolArrayToBitGroups(finalBoolArr)
+			missedInfo.WindowSize = window
+			// TODO: set missed blocks later
+			// k.SetValidatorMissedBlocks(ctx, writeInfo.ConsAddr, missedInfo)
+		case missedInfo.WindowSize > window:
+			// missed block array too long, we need to trim
+			// we need to keep the last N blocks prior to the validator index offset (wrapping around backwards if necessary)
```
yzang2019, line 78: IMO Shrinking window size is a risky operation that might cause lot of validators to be jailed at the same time when window size change if here we decided to keep the history.

My recommendation (not sure if make sense or not) is to keep things simple, let's just drop all the history when the window size is shrinked, recreate a new empty array list to start with. This window size change should be pretty rare to happen and it's actually safer to if we missed to punish validators than we suddenly started to jail a bunch of validators.

x/slashing/keeper/signing_info.go
```diff
@@ -72,6 +72,62 @@ func (k Keeper) SetValidatorMissedBlocks(ctx sdk.Context, address sdk.ConsAddres
 	store.Set(types.ValidatorMissedBlockBitArrayKey(address), bz)
 }
 
+// Get a boolean representing whether a validator missed a block with a specific index offset
+func (k Keeper) GetValidatorMissedBlockBitFromArray(bitGroupArray []uint64, index int64) bool {
```
yzang2019, line 76: This could even be a simple util/helper function.

GetBoolean (bitArray, index)
SetBoolean(bitArray, index)

x/slashing/keeper/signing_info.go
```diff
@@ -72,6 +72,62 @@ func (k Keeper) SetValidatorMissedBlocks(ctx sdk.Context, address sdk.ConsAddres
...
+		bitGroupArray[indexKey] |= indexMask
+	} else {
+		// set bit to 0 by AND NOTing with the specific position as 1
+		bitGroupArray[indexKey] &^= indexMask
+	}
+	// set after updating the position
+	return bitGroupArray
+}
+
+func (k Keeper) ParseBitGroupsToBoolArray(bitGroups []uint64, window int64) []bool {
```
yzang2019, line 107: I think we should avoid parsing bit array to boolean array and vice versa if possible

x/slashing/keeper/infractions.go
```diff
@@ -36,30 +36,105 @@ func (k Keeper) HandleValidatorSignatureConcurrent(ctx sdk.Context, addr cryptot
...
+			signInfo.MissedBlocksCounter = int64(newMissedCount)
+			missedInfo.WindowSize = window
+			// TODO: set missed blocks later
+			// k.SetValidatorMissedBlocks(ctx, writeInfo.ConsAddr, missedInfo)
 		}
 	}
-	// update Missed Heights by excluding heights outside of the window
-	missedInfo.MissedHeights = missedInfo.MissedHeights[numberEvicted:]
+	// bump index offset after performing potential resizing
+	signInfo.IndexOffset++
```
yzang2019, line 105: let's change this to circular index

proto/cosmos/slashing/v1beta1/genesis.proto
```diff
@@ -22,14 +22,30 @@ message GenesisState {
       [(gogoproto.moretags) = "yaml:\"missed_blocks\"", (gogoproto.nullable) = false];
 }
 
+// GenesisState defines the slashing module's genesis state.
+message GenesisStateLegacyMissingHeights {
+  // params defines all the paramaters of related to deposit.
```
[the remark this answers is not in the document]
udpatil: Yeah this was copied from a previous one maybe that one is incorrect as well, I'll check

proto/cosmos/slashing/v1beta1/slashing.proto
```diff
@@ -50,12 +50,21 @@ message ValidatorSigningInfoLegacyV43 {
 }
 
 // Stores a sliding window of the last `signed_blocks_window` blocks indicating whether the validator missed the block
-message ValidatorMissedBlockArray {
+message ValidatorMissedBlockArrayLegacyMissedHeights {
```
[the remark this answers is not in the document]
udpatil: Yeah we need the old schema for loading the data from store for the migration

x/slashing/keeper/infractions.go
```diff
@@ -36,30 +36,105 @@ func (k Keeper) HandleValidatorSignatureConcurrent(ctx sdk.Context, addr cryptot
 
 	window := k.SignedBlocksWindow(ctx)
 
+	index := signInfo.IndexOffset % k.SignedBlocksWindow(ctx)
+
 	missedInfo, found = k.GetValidatorMissedBlocks(ctx, consAddr)
+
 	if !found {
+		arrLen := window / 64
```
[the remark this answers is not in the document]
udpatil: Ah yeah good point, will fix

x/slashing/keeper/infractions.go
```diff
@@ -36,30 +36,105 @@ func (k Keeper) HandleValidatorSignatureConcurrent(ctx sdk.Context, addr cryptot
...
-	cutoff := height - window
-	// go through and increment number evicted as necessary
-	numberEvicted := 0
-	for _, missedHeight := range missedInfo.MissedHeights {
-		if missedHeight <= cutoff {
-			numberEvicted += 1
-		} else {
-			break
+	// TODO:  resize if necessary?
+	if found && missedInfo.WindowSize != window {
```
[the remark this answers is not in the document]
udpatil: Yup will do

x/slashing/keeper/infractions.go
udpatil: Yeah this is reasonable, although ideally vals shouldn't be jailed because we update missed counters to exclude missed blocks from outside the window

x/slashing/keeper/signing_info.go
udpatil: agreed, this is only done when expanding the missed block array (eg. only when they have param changes)

x/slashing/keeper/signing_info.go
```diff
@@ -72,6 +72,58 @@ func (k Keeper) SetValidatorMissedBlocks(ctx sdk.Context, address sdk.ConsAddres
...
+	}
+	// shift 1 by the indexShift value to generate bit mask (to index into the bitGroup)
+	indexMask := uint64(1) << indexShift
+	// apply the mask and if the value at that `indexShift` is 1 (indicating miss), then the value would be non-zero
+	missed := (bitGroupArray[indexKey] & indexMask) != 0
+	return missed
+}
+
+// Set the missed value for whether a validator missed a block
+func (k Keeper) SetGetBooleanInBitGroups(bitGroupArray []uint64, index int64, missed bool) []uint64 {
```
philipsu522, line 91: nit - can we rename this to just `SetBooleanInBitGroups`, or am i missing something with the "Get" part?

x/slashing/keeper/infractions.go
```diff
@@ -36,30 +36,38 @@ func (k Keeper) HandleValidatorSignatureConcurrent(ctx sdk.Context, addr cryptot
 
 	window := k.SignedBlocksWindow(ctx)
 
+	index := signInfo.IndexOffset
+
 	missedInfo, found = k.GetValidatorMissedBlocks(ctx, consAddr)
+
 	if !found {
+		arrLen := (window + 63) / 64
```
philipsu522, line 44: would recommend using a static final variable here, i.le.
```
		arrLen := (window + INT_64_NUM_BITS - 1) / INT_64_NUM_BITS
```

x/slashing/keeper/infractions.go
```diff
@@ -36,30 +36,38 @@ func (k Keeper) HandleValidatorSignatureConcurrent(ctx sdk.Context, addr cryptot
...
-	missedInfo.MissedHeights = missedInfo.MissedHeights[numberEvicted:]
+	previous := k.GetBooleanFromBitGroups(missedInfo.MissedBlocks, index)
 	missed := !signed
-	if missed {
-		missedInfo.MissedHeights = append(missedInfo.MissedHeights, height-1)
+	switch {
+	case !previous && missed:
+		// Array value has changed from not missed to missed, increment counter
+		signInfo.MissedBlocksCounter++
+		missedInfo.MissedBlocks = k.SetGetBooleanInBitGroups(missedInfo.MissedBlocks, index, true)
```
philipsu522, line 60: nit - last param could just be `missed`

x/slashing/keeper/infractions.go
```diff
@@ -36,30 +36,38 @@ func (k Keeper) HandleValidatorSignatureConcurrent(ctx sdk.Context, addr cryptot
...
+	case previous && !missed:
+		// Array value has changed from missed to not missed, decrement counter
+		signInfo.MissedBlocksCounter--
+		missedInfo.MissedBlocks = k.SetGetBooleanInBitGroups(missedInfo.MissedBlocks, index, false)
```
philipsu522, line 64: same here, and can move this out of switch/case

x/slashing/keeper/infractions.go
```diff
@@ -142,3 +151,29 @@ func (k Keeper) SlashJailAndUpdateSigningInfo(ctx sdk.Context, consAddr sdk.Cons
...
+	// we need to resize the missed block array AND update the signing info accordingly
+	switch {
+	case missedInfo.WindowSize < window:
+		// missed block array too short, lets expand it
+		boolArray := k.ParseBitGroupsToBoolArray(missedInfo.MissedBlocks, missedInfo.WindowSize)
+		newArray := make([]bool, window)
+		copy(newArray[0:index], boolArray[0:index])
+		if index+1 < missedInfo.WindowSize {
+			// insert `0`s corresponding to the difference between the new window size and old window size
+			copy(newArray[index+(window-missedInfo.WindowSize):], boolArray[index:])
```
philipsu522, line 165: would it just be easier to have newArray filled with `0`'s and then update [0:index] in both? I guess i'm confused what the last line here is doing

x/slashing/keeper/migrations.go
```diff
@@ -118,7 +123,92 @@ func (m Migrator) Migrate2to3(ctx sdk.Context) error {
...
+			StartHeight:         oldInfo.StartHeight,
+			IndexOffset:         index,
+			JailedUntil:         oldInfo.JailedUntil,
+			Tombstoned:          oldInfo.Tombstoned,
+			MissedBlocksCounter: oldInfo.MissedBlocksCounter,
+		}
+		newSignInfoKeys = append(newSignInfoKeys, signInfoIter.Key())
+		newSignInfoVals = append(newSignInfoVals, newInfo)
+	}
+	signInfoIter.Close()
```
philipsu522, line 162: we're closing the iterator twice - is that intentional?

x/slashing/keeper/migrations.go
```diff
@@ -118,7 +123,92 @@ func (m Migrator) Migrate2to3(ctx sdk.Context) error {
...
+
+		valMissedMap[consAddr.String()] = types.ValidatorMissedBlockArray{
+			Address:      missedInfo.Address,
+			MissedBlocks: m.keeper.ParseBoolArrayToBitGroups(newBoolArray),
+			WindowSize:   window,
+		}
+	}
+
+	ctx.Logger().Info("Writing new validator missed blocks infos")
+	for key, missedBlockArray := range valMissedMap {
```
philipsu522, line 205: this should be fine, but we should sort this anyways to avoid any source of nondeterm

x/slashing/keeper/signing_info.go
```diff
@@ -72,6 +72,58 @@ func (k Keeper) SetValidatorMissedBlocks(ctx sdk.Context, address sdk.ConsAddres
 	store.Set(types.ValidatorMissedBlockBitArrayKey(address), bz)
 }
 
+// Get a boolean representing whether a validator missed a block with a specific index offset
+func (k Keeper) GetBooleanFromBitGroups(bitGroupArray []uint64, index int64) bool {
+	// convert the index into indexKey + indexShift
+	indexKey := index / 64
```
philipsu522, line 78: same nit comment re. the static final variable

philipsu522 (approved):
mostly minor comments, overall lgtm!

x/slashing/keeper/signing_info.go
udpatil: nah this is a typo

x/slashing/keeper/infractions.go
udpatil: we need the previous value to calculate whether the missed counter needs incrementing or decrementing. However we can move the `setBooleanInBitGroups` out of switch case

x/slashing/keeper/infractions.go
udpatil: we need to update 0:index and then insert N corresponding to the number of blocks added to the signing window, and THEN include the rest of the array. This is because we want to update the X oldest value (x=signwindow).

Example
```
[0, 1, 1, 0] index = 2 (resize from 4 -> 6) 
because the above index = 2 is the 4th oldest value (we know this because it was about to be updated), it should only be updated after updating the 6th and 5th oldest values.
therefore we result with the expanded array

[0, 1, 0, 0, 1, 0]
in this, index 2 and 3 represent the expanded values and we update index = 2 because that represents the 6th oldest block
```

Lmk if this doesn't make sense or if i'm mistunderstanding your suggested modification

x/slashing/keeper/migrations.go
udpatil: yeah, @yzang2019 said this is fine. We're closing this one here to make sure the next one being opened is done safely, and the defer for close is in case theres an error while iterating

philipsu522 (approved):

PR #218, opened 2023-04-14
Restart node instead of panicing and exiting 

BrandonWeng:
Currently the node will only switch into blocksync mode if the whole process is restarted, this change aims to add a detection go routine and restart the node if it's behind configurable 

Artificially introduced latency so that the node would fall behind, and see that it self restarted node 
image

image

### DIFF
--- a/baseapp/baseapp.go
+++ b/baseapp/baseapp.go
@@ -1092,3 +1092,10 @@ func (app *BaseApp) startCompactionRoutine(db dbm.DB) {
 		}
 	}()
 }
+
+func (app *BaseApp) Close() error {
+	if err := app.appStore.db.Close(); err != nil {
+		return err
+	}
+	return app.snapshotManager.Close()
+}
--- a/go.mod
+++ b/go.mod
@@ -41,6 +41,7 @@ require (
 	github.com/spf13/pflag v1.0.5
 	github.com/spf13/viper v1.12.0
 	github.com/stretchr/testify v1.8.1
+	github.com/syndtr/goleveldb v1.0.1-0.20210819022825-2ae1ddf74ef7
 	github.com/tendermint/btcd v0.1.1
 	github.com/tendermint/crypto v0.0.0-20191022145703-50d29ede1e15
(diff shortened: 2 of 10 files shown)


BrandonWeng:
```
Error: stat foobar/config/genesis.json: no such file or directory
Usage:
  export [flags]

Flags:
      --chain-id string              Chain ID
      --for-zero-height              Export state to start at height zero (perform preproccessing)
      --height int                   Export state from a particular height (-1 means latest height) (default -1)
  -h, --help                         help for export
      --home string                  The application home directory (default "/tmp/TestExportCmd_HomeDir378377008/001")
      --jail-allowed-addrs strings   Comma-separated list of operator addresses of jailed validators to unjail
``` 
I think the test failures are unrelated

yzang2019:
The test failures seems to be new, I think we should aim to fix them or at least figure out why they are failing. If they fail due to flakyness, then that's probably fine

philipsu522 (approved):
Looks good. The only thing I'm slightly concerned about is that if we forcefully shut down the process (which I don't think it does) in the middle of a write to the database. This could cause some integrity issues and ultimately bork a machine. Can we:
1. make sure we pause and wait for all graceful shutdown routines
2. let it run repeatedly on a node for a while?

BrandonWeng:
> Looks good. The only thing I'm slightly concerned about is that if we forcefully shut down the process (which I don't think it does) in the middle of a write to the database. This could cause some integrity issues and ultimately bork a machine. Can we:
> 
> 1. make sure we pause and wait for all graceful shutdown routines
> 2. let it run repeatedly on a node for a while?

1. In the StartInProcess call we actually should be gracefully shutting down the process whenever a restart/sig is sent (including database connections).  https://github.com/sei-protocol/sei-cosmos/pull/218/files#diff-97289af2ab66016f168df6fccac70626d125f170e5cf0841fcdc8fa624999608R497-R521
2. Yeah I can spin up a chain with this change and let it run for a while (with a couple of the nodes with smaller instances so they fall behind)

PR #219, opened 2023-04-17
Add snapshot tendermint command

codchen:
Add command

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -41,6 +41,7 @@ require (
 	github.com/spf13/pflag v1.0.5
 	github.com/spf13/viper v1.12.0
 	github.com/stretchr/testify v1.8.1
+	github.com/syndtr/goleveldb v1.0.1-0.20210819022825-2ae1ddf74ef7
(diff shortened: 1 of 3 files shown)


baseapp/baseapp.go
```diff
@@ -1099,3 +1101,18 @@ func (app *BaseApp) Close() error {
 	}
 	return app.snapshotManager.Close()
 }
+
+func (app *BaseApp) ReloadDB() error {
+	if err := app.db.Close(); err != nil {
+		return err
+	}
+	rootDir := "/root/.sei"
```
yzang2019, line 1109: This directory isn't always the same for all applications right? I believe we can customize the directory? Shall we pass it from somewhere instead of hard coding it?

codchen: oh good catch I hardcoded it for testing. Updated to be config-based

philipsu522 (approved):

yzang2019 (approved):

PR #220, opened 2023-04-18
Add admin field to base vesting account

udpatil:
This allows for configuring vesting accounts with an admin

Added unit tests and tested with local sei


philipsu522 (approved):
is a store migration needed?

x/auth/vesting/client/cli/tx.go
```diff
@@ -44,7 +45,7 @@ func NewMsgCreateVestingAccountCmd() *cobra.Command {
 account can either be a delayed or continuous vesting account, which is determined
 by the '--delayed' flag. All vesting accouts created will have their start time
 set by the committed block's time. The end_time must be provided as a UNIX epoch
-timestamp.`,
+timestamp. You can also optionally configure the 'admin' field using the flag '--admin {addr}. This admin will be able to perform some administrative actions on the vesting account if set.`,
```
BrandonWeng, line 48: Looks like the vesting submodule doesn't have a query handler, is there a way to query or display the account admin from the CLI/RPC endpoint or add it a query handler or display it in `seid q auth account(s)`?

udpatil:
> is a store migration needed?

yeah good point, I'll add one in

x/auth/vesting/client/cli/tx.go
udpatil: it's returned when querying the actual account `seid q auth account $addr`

udpatil:
actually, @psu It seems like maybe we don't need a store migration, since when we query accounts created prior to the vesting account change, they show admin as `""` which is correct

PR #225, opened 2023-04-21
Add latest_version command to get the latest version

codchen:
Usage: `seid latest_version`
Output: <latest application DB version>

local sei

### DIFF
--- a/server/version.go
+++ b/server/version.go
@@ -39,7 +39,6 @@ func LatestVersionCmd(defaultNodeHome string) *cobra.Command {
 	}
 
 	cmd.Flags().String(flags.FlagHome, defaultNodeHome, "The application home directory")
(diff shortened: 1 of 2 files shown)


server/version.go
```diff
@@ -0,0 +1,45 @@
...
+				return err
+			}
+			fmt.Println(rootmulti.GetLatestVersion(db))
+
+			return nil
+		},
+	}
+
+	cmd.Flags().String(flags.FlagHome, defaultNodeHome, "The application home directory")
+	cmd.Flags().String(FlagChainID, "", "Chain ID")
```
yzang2019, line 42: why do we need ChainId flag? It seems this is not used in the above logic

codchen: yeah good point. Removed

yzang2019 (approved):

PR #235, opened 2023-05-02
Fix fee check for simulate mode

codchen:
We check if the provided fee meets validators' minimum limit in CheckTx to prevent spamming, but simulate also uses CheckTx's state, and since simulate is run with gas limit set to `auto`, the check here shouldn't apply.

tested with unit test in sei-chain


philipsu522 (approved):
@besated tested and verified. merging

PR #238, opened 2023-05-11
Refactor tracing to sei-cosmos

udpatil:
This refactors tracing from sei-chain app to sei-cosmos baseapp, allowing us to add tracing to sei-cosmos level logic
tested locally


server/start.go
```diff
@@ -52,7 +53,6 @@ const (
 	FlagInterBlockCache    = "inter-block-cache"
 	FlagUnsafeSkipUpgrades = "unsafe-skip-upgrades"
 	FlagTrace              = "trace"
-	FlagTracing            = "tracing"
```
BrandonWeng, line 55: nit, should we just keep it here to be consistent with all the other Flag constants?

baseapp/baseapp.go
```diff
@@ -817,6 +843,19 @@ func (app *BaseApp) runTx(ctx sdk.Context, mode runTxMode, txBytes []byte) (gInf
...
+	if tracer != nil {
+		spanCtx, span := (*tracer).Start(spanCtx, "RunTx")
+		span.SetAttributes(attribute.String("txHash", fmt.Sprintf("%X", sha256.Sum256(txBytes))))
+		// update ctx span and tracer in case of further child tracing
+		ctx = ctx.WithSpanAndTracer(spanCtx, tracer)
+		defer span.End()
+	} else {
+		// this is unexpected, but we can just make a normal span
+		_, span := app.TracingInfo.Start("RunTx")
+		defer span.End()
```
BrandonWeng, line 857: q: why is this unexpected? :o

udpatil: Because we will always be passing in a parent span for the initial delivertx call (see sei-chain abci.go)

server/start.go
udpatil: Putting it here requires an import of this file in baseapp.go which complains about an import cycle

BrandonWeng (approved):

philipsu522 (approved):

yzang2019 (approved):

PR #240, opened 2023-05-12
No Slashing by default

BrandonWeng:
As discussed we will be disabling slashing by default and we'll be updating our existing chains 

Updated unite tests

### DIFF
--- a/x/slashing/keeper/keeper_test.go
+++ b/x/slashing/keeper/keeper_test.go
@@ -142,6 +142,11 @@ func TestHandleAlreadyJailed(t *testing.T) {
 	params.SignedBlocksWindow = 1000
 	app.SlashingKeeper.SetParams(ctx, params)
 
+	slashingParams := app.SlashingKeeper.GetParams(ctx)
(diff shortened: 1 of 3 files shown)


x/slashing/keeper/keeper_test.go
```diff
@@ -163,16 +163,15 @@ func TestHandleAlreadyJailed(t *testing.T) {
 	require.Equal(t, stakingtypes.Unbonding, validator.GetStatus())
 
 	// validator should have been slashed
-	resultingTokens := amt.Sub(app.StakingKeeper.TokensFromConsensusPower(ctx, 1))
-	require.Equal(t, resultingTokens, validator.GetTokens())
+	require.Equal(t, amt, validator.GetTokens())
```
udpatil, line 166: i dont think we should modify the tests we would like to still establish that the slashing works as expected if needed in the future, lets set the slashing percentage to the old defaults in the tests prior to test execution

udpatil (approved):

PR #253, opened 2023-05-26
Enhance the iavl tree metric reliability

yzang2019:
**Problem**:
Currently, we only expose a single metric point after each snapshot is done. However, in atlantic-2, each snapshot could take multiple hours to finish, this makes the metric itself very unreliable. 

**Solution**:
Instead of setting gauge every time after snapshot is done, we create a new goroutine to keep setting the gauge every 60 seconds, and we only update the value after the snapshot is done

Tested in atlantic-2 state syncer node

### DIFF
--- a/store/rootmulti/store.go
+++ b/store/rootmulti/store.go
@@ -8,6 +8,8 @@ import (
 	"sort"
 	"strings"
 	"sync"
+	"sync/atomic"
+	"time"
 
 	"github.com/armon/go-metrics"
 	"github.com/cosmos/cosmos-sdk/telemetry"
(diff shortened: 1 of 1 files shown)


BrandonWeng (commented):
let me take this on mondray, there’s a way for prometheus to just  last value if it’s not update. that’s how comos exporter does it

philipsu522 (approved):

yzang2019:
> let me take this on mondray, there’s a way for prometheus to just last value if it’s not update. that’s how comos exporter does it

Thanks a lot! Let me know if you have any updates on this

yzang2019:
Switching to a better approach: https://github.com/sei-protocol/sei-cosmos/pull/257

PR #256, opened 2023-05-29
add proper ante deps for feegrant

udpatil:
This fixes the generated dependencies for txs that use fee grants. The deductFee antehandler previously incorrectly declared the dependencies, causing an invalid concurrency remediation in a block with a tx using a fee grant.

Tested in localsei to verify txs paying via fee grant work concurrently.


philipsu522 (approved):

x/auth/ante/fee.go
```diff
@@ -57,31 +59,53 @@ func (d DeductFeeDecorator) AnteDeps(txDeps []sdkacltypes.AccessOperation, tx sd
 			[]sdkacltypes.AccessOperation{
 				{
 					AccessType:         sdkacltypes.AccessType_READ,
-					ResourceType:       sdkacltypes.ResourceType_KV_BANK_BALANCES,
-					IdentifierTemplate: hex.EncodeToString(banktypes.CreateAccountBalancesPrefix(feeTx.FeePayer())),
+					ResourceType:       sdkacltypes.ResourceType_KV_AUTH_ADDRESS_STORE,
```
BrandonWeng, line 62: we have unit tests for message types to issues, IRRC helped a bunch with Tony's refactor but it seems like it caught the ante handler issue. Could we add some unit tests for the antehandler code paths too?

udpatil: yup, will do

udpatil: Added unit tests

PR #258, opened 2023-05-30
Add checkTx parsedLogs

LCyson:
This PR parses broadcastTx log so that it can be correctly shown in the response.log:

w/ this PR:
```
(base) ➜  sei-chain git:(master) ✗ seid tx bank send admin sei1txqs5kqjx5g33syy6mhtgq7z77wp2kf8r72qf6 100000sei -b block -y --chain-id sei-chain --fees=1000000usei --gas=500000 --sequence 10
code: 32
codespace: sdk
data: ""
events: []
gas_used: "0"
gas_wanted: "0"
height: "0"
info: ""
logs:
- events: []
  log: 'account sequence mismatch, expected 1, got 10: incorrect account sequence'
  msg_index: 0
raw_log: ""
timestamp: ""
tx: null
txhash: E71A5A9C34B5E09BE4116C77C0AA7FA63722FAA17A20FE229A8579E36BC5A318
```

w/o this PR:
```
...
logs: ""
raw_log: ""
...
```

- unit tests
- e2e test

### DIFF
--- a/types/result.go
+++ b/types/result.go
@@ -115,6 +115,7 @@ func newTxResponseCheckTx(res *ctypes.ResultBroadcastTxCommit) *TxResponse {
 		Codespace: res.CheckTx.Codespace,
 		Code:      res.CheckTx.Code,
 		Data:      strings.ToUpper(hex.EncodeToString(res.CheckTx.Data)),
+		RawLog:    res.CheckTx.Log,
 		Logs:      parsedLogs,
 		GasWanted: res.CheckTx.GasWanted,
 	}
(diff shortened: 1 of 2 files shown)


yzang2019 (approved):

philipsu522 (approved):

types/result.go
```diff
@@ -107,13 +107,15 @@ func newTxResponseCheckTx(res *ctypes.ResultBroadcastTxCommit) *TxResponse {
...
+	parsedLogs, _ := ParseABCILogs(res.CheckTx.Log)
+
 	return &TxResponse{
 		Height:    res.Height,
 		TxHash:    txHash,
 		Codespace: res.CheckTx.Codespace,
 		Code:      res.CheckTx.Code,
 		Data:      strings.ToUpper(hex.EncodeToString(res.CheckTx.Data)),
-		RawLog:    res.CheckTx.Log,
+		Logs:      parsedLogs,
```
philipsu522, line 118: is it possible to add a new field rather than replacing it, for backwards compatibility (in case anyone uses raw log)

philipsu522 (changes requested):
added comment

types/result.go
LCyson: we added the rawLog in the last patch so it shouldn't have been used by anybody (was used for logging purposes), but fine to leave both there

PR #263, opened 2023-06-01
add more unit tests for global fee param 

BrandonWeng:
adding more tests. cases for global fee param for multiple denoms


philipsu522 (approved):
what a test 🔥

PR #265, opened 2023-06-01
Change dec coin amount of to iterative too

BrandonWeng:
This changes amount of for dec coins to be iterative too it dangerously assumes that the coins are in sorted order and returns 0 often if it's not.

Added a unit test for this, 

it would fail with the old logic

image

### DIFF
--- a/x/auth/ante/validator_tx_fee.go
+++ b/x/auth/ante/validator_tx_fee.go
@@ -24,7 +24,7 @@ func CheckTxFeeWithValidatorMinGasPrices(ctx sdk.Context, tx sdk.Tx, simulate bo
 	// is only ran on check tx.
 	if ctx.IsCheckTx() && !simulate {
 		feeParams := paramsKeeper.GetFeesParams(ctx)
-		minGasPrices := GetMinimumGasPricesWanted(feeParams.GetGlobalMinimumGasPrices(), ctx.MinGasPrices())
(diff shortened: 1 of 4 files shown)


x/auth/ante/validator_tx_fee.go
```diff
@@ -51,8 +51,8 @@ func CheckTxFeeWithValidatorMinGasPrices(ctx sdk.Context, tx sdk.Tx, simulate bo
 	return feeCoins, priority, nil
 }
 
-func GetMinimumGasPricesWanted(globalMinimumGasPrices, validatorMinimumGasPrices sdk.DecCoins) sdk.DecCoins {
-	return globalMinimumGasPrices.UnionMax(validatorMinimumGasPrices)
+func GetMinimumGasPricesWantedSorted(globalMinimumGasPrices, validatorMinimumGasPrices sdk.DecCoins) sdk.DecCoins {
+	return globalMinimumGasPrices.UnionMax(validatorMinimumGasPrices).Sort()
```
BrandonWeng, line 55: I think this was the reason why when @udpatil tested sorted ordered local min gas prices it was still failing, when combined it would not preserve the sorted order. 

Not really needed with us switching to iterative ordering but this is consistent with what Gaia does: https://github.com/cosmos/gaia/blob/main/x/globalfee/ante/fee_utils.go#LL51C1-L51C1

philipsu522 (approved):

PR #266, opened 2023-06-02
Remove deferred withdrawals and add saturating sub with tests

udpatil:
This does several things:
1. Removes the deferred withdrawals and deferred mints/burns.
2. Adds a saturating sub as an alternative to safesub. We can use this and pass ing the subUnlockedCoins to atomically perform a saturating sub against the deferred balance first and then against the underlying balance using subUnlockedCoins.
3. Added a partition function for the coins to separate positive and negative values to separate remainders for the saturating sub
4. refactors the bank keepers to perform saturating subtraction instead of safesub

Unit tests... LOTS of unit tests

### DIFF
--- a/types/coin.go
+++ b/types/coin.go
@@ -390,6 +390,19 @@ func (coins Coins) SafeSub(coinsB Coins) (Coins, bool) {
 	return diff, diff.IsAnyNegative()
 }
 
+// PartitionSigned separates the coins between those with positive values and those with negative values,
+// and returns both positive and negative coins. This also discards any coins with zero values (not returned in either group).
+func (coins Coins) PartitionSigned() (positives Coins, negatives Coins) {
+	for _, coin := range coins {
+		if coin.Amount.IsPositive() {
+			positives = append(positives, coin)
+		} else if coin.Amount.IsNegative() {
+			negatives = append(negatives, coin)
+		}
+	}
+	return positives, negatives
+}
+
 // Max takes two valid Coins inputs and returns a valid Coins result
 // where for every denom D, AmountOf(D) of the result is the maximum
 // of AmountOf(D) of the inputs.  Note that the result might be not
--- a/types/context_cache.go
+++ b/types/context_cache.go
@@ -73,6 +73,8 @@ func (c *ContextMemCache) SafeSubDeferredSends(moduleAccount string, amount Coin
 }
 
 func (c *ContextMemCache) RangeOnDeferredSendsAndDelete(apply func(recipient string, amount Coins)) {
+	c.deferredBankOpsLock.Lock()
+	defer c.deferredBankOpsLock.Unlock()
 	c.deferredSends.RangeAndRemove(apply)
 }
 
--- a/types/context_cache.go
+++ b/types/context_cache.go
@@ -68,6 +68,8 @@ func (c *ContextMemCache) SafeSubDeferredSends(moduleAccount string, amount Coin
 	if !amount.IsValid() {
 		panic(sdkerrors.Wrap(sdkerrors.ErrInvalidCoins, amount.String()))
 	}
+	c.deferredBankOpsLock.Lock()
+	defer c.deferredBankOpsLock.Unlock()
 
 	return c.deferredSends.SafeSub(moduleAccount, amount)
 }
--- a/types/context_cache.go
+++ b/types/context_cache.go
@@ -40,26 +40,28 @@ func (c *ContextMemCache) SaturatingSubDeferredSends(moduleAccount string, amoun
 	if !amount.IsValid() {
 		panic(sdkerrors.Wrap(sdkerrors.ErrInvalidCoins, amount.String()))
 	}
+	c.deferredBankOpsLock.Lock()
+	defer c.deferredBankOpsLock.Unlock()
+
 	originalBalance, ok := c.deferredSends.Get(moduleAccount)
-	if ok {
-		// found a balance, try to perform the logic
-		remainder := c.deferredSends.SaturatingSub(moduleAccount, amount)
-		if remainder.IsZero() {
-			// no remainder, return now
-			return nil
-		}
-		err := remainderHandler(remainder)
-		if err != nil {
-			// revert the map subtraction and the bubble up error
-			c.deferredSends.Set(moduleAccount, originalBalance)
-		}
-		return err
-	} else {
+	if !ok {
 		// run the remainder handler on the full amount
 		err := remainderHandler(amount)
 		// no need for revert because we dont even attempt a saturating sub, bubble up error if applicable
 		return err
 	}
+	// found a balance, try to perform the logic
+	remainder := c.deferredSends.SaturatingSub(moduleAccount, amount)
+	if remainder.IsZero() {
+		// no remainder, return now
+		return nil
+	}
+	err := remainderHandler(remainder)
+	if err != nil {
+		// revert the map subtraction and the bubble up error
+		c.deferredSends.Set(moduleAccount, originalBalance)
+	}
+	return err
 }
 
 func (c *ContextMemCache) SafeSubDeferredSends(moduleAccount string, amount Coins) bool {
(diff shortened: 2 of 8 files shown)


types/context_cache.go
```diff
@@ -37,73 +31,51 @@ func (c *ContextMemCache) UpsertDeferredSends(moduleAccount string, amount Coins
...
-	// Separate locks needed for all mappings - atmoic transaction needed
-	c.deferredBankOpsLock.Lock()
-	defer c.deferredBankOpsLock.Unlock()
-
-	// If there's already a pending deposit then subtract it from that amount first
-	// or else add it to the deferredWithdrawals mapping
-	ok := c.deferredSends.SafeSub(moduleAccount, amount)
-	if !ok {
-		c.deferredWithdrawals.UpsertMapping(moduleAccount, amount)
+	originalBalance, ok := c.deferredSends.Get(moduleAccount)
```
BrandonWeng, line 43: I think we need to lock/unlock here or else we'll have race conditions:
```
	c.deferredBankOpsLock.Lock()
	defer c.deferredBankOpsLock.Unlock()
```

types/context_cache.go
```diff
@@ -37,73 +31,51 @@ func (c *ContextMemCache) UpsertDeferredSends(moduleAccount string, amount Coins
...
+			// revert the map subtraction and the bubble up error
+			c.deferredSends.Set(moduleAccount, originalBalance)
+		}
+		return err
+	} else {
+		// run the remainder handler on the full amount
+		err := remainderHandler(amount)
+		// no need for revert because we dont even attempt a saturating sub, bubble up error if applicable
+		return err
 	}
```
BrandonWeng, line 62:
```suggestion
	if !ok {
		// run the remainder handler on the full amount
		err := remainderHandler(amount)
		// no need for revert because we dont even attempt a saturating sub, bubble up error if applicable
		return err
	} 
        // found a balance, try to perform the logic
        remainder := c.deferredSends.SaturatingSub(moduleAccount, amount)
        if remainder.IsZero() {
	        // no remainder, return now
	        return nil
        }
        err := remainderHandler(remainder)
        if err != nil {
	        // revert the map subtraction and the bubble up error
	        c.deferredSends.Set(moduleAccount, originalBalance)
        }
        return err

	
```

types/deferred_bank_operations.go
```diff
@@ -7,23 +7,70 @@ import (
...
+	defer m.mappingLock.Unlock()
+
+	m.deferredOperations[moduleAccount] = amount
+}
+
+// SaturatingSub will subtract the given amount from the module account as long as the resulting balance is positive or zero.
+// If there would be a remainder (eg. negative balance after subtraction), then it would subtract the full balance in the map
+//
+//	and then return the remainder that was unable to be subtracted.
+func (m *DeferredBankOperationMapping) SaturatingSub(moduleAccount string, amount Coins) Coins {
```
BrandonWeng, line 43: nit, i think the formatting here is a bit weird

BrandonWeng (commented):
thanks for the change, could we turn parallel tests on? One thing that I'm concerned about the unit tests not catching is any race conditionals with parallel TX turned on

```
func TestMyFunction(t *testing.T) {
    t.Parallel()
    //your test code here

```

types/context_cache.go
udpatil: good catch

types/context_cache.go
```diff
@@ -37,73 +31,53 @@ func (c *ContextMemCache) UpsertDeferredSends(moduleAccount string, amount Coins
...
+		c.deferredSends.Set(moduleAccount, originalBalance)
+	}
+	return err
 }
 
-// This inserts or updates an entry for a module account for a deferred withdrawal.
-// This should be performed AFTER checking for a sufficient balance in the underlying bank balances for that module account.
-// Additionally, this does not attempt to do any safe subtraction from pending sends, so if that behavior is preferred, it will need to be checked separately.
-func (c *ContextMemCache) UpsertDeferredWithdrawalsNoSafeSub(moduleAccount string, amount Coins) error {
+func (c *ContextMemCache) SafeSubDeferredSends(moduleAccount string, amount Coins) bool {
```
philipsu522, line 67: from the PR description, it seems like this is not meant to be called publicly. Can we change this to start w/ lower case so it's not exported?

types/context_cache.go
```diff
@@ -37,73 +31,53 @@ func (c *ContextMemCache) UpsertDeferredSends(moduleAccount string, amount Coins
...
-	if !ok {
-		c.deferredSends.UpsertMapping(moduleAccount, amount)
-	}
+	c.deferredSends.UpsertMapping(moduleAccount, amount)
 	return nil
 }
 
-func (c *ContextMemCache) SafeSubDeferredSends(moduleAccount string, amount Coins) bool {
+// This will perform a saturating sub against the deferred module balance atomically. This means that it will subtract any balances that it is able to, and then call the passed in `remainder handler` to ensure the remaining balance can be safely subtracted as well. If this remainderHandler returns an error, we will revert the saturating sub to ensure atomicity of the subtraction across the multiple balances
+func (c *ContextMemCache) SaturatingSubDeferredSends(moduleAccount string, amount Coins, remainderHandler func(amount Coins) error) error {
```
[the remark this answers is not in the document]
udpatil: I used saturating to mean it saturates from the deferred balance before subtracting from bank balances, but if we have a better name then we can use that. `AtomicSubDeferredSends` would work but doesn't describe that it spills over to bank balance for the remainder

types/coin.go
```diff
@@ -390,6 +390,19 @@ func (coins Coins) SafeSub(coinsB Coins) (Coins, bool) {
 	return diff, diff.IsAnyNegative()
 }
 
+// PartitionSigned separates the coins between those with positive values and those with negative values,
+// and returns both positive and negative coins. This also discards any coins with zero values (not returned in either group).
+func (coins Coins) PartitionSigned() (positives Coins, negatives Coins) {
+	for _, coin := range coins {
```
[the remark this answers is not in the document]
udpatil: zero values are removed (in some other coin logic, zero values are removed too, just explicitly using removeZeroCoins)

types/context_cache_test.go
```diff
@@ -0,0 +1,279 @@
...
+
+func (s *contextCacheTestSuite) SetupSuite() {
+	s.T().Parallel()
+	s.contextCache = *sdk.NewContextMemCache()
+}
+
+func (s *contextCacheTestSuite) TestDeferredSendUpserts() {
+	s.contextCache.UpsertDeferredSends("module1", sdk.NewCoins(
+		sdk.NewInt64Coin("denom1", 100),
+		sdk.NewInt64Coin("denom2", 20),
```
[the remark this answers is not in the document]
udpatil: I dont think negative coins are allowed here, i believe most callsites of this function have behavior disallowing negative coins. Additionally, a lot of coin login panics upon receiving negative coins.

types/context_cache.go
udpatil: do we still want to export it so that the unit tests for it are still able to run?

BrandonWeng: do we need mutex lock on this? looks like we're doing a write

types/context_cache.go
```diff
@@ -37,73 +31,53 @@ func (c *ContextMemCache) UpsertDeferredSends(moduleAccount string, amount Coins
...
-func (c *ContextMemCache) RangeOnDeferredWithdrawalsAndDelete(apply func(recipient string, amount Coins)) {
-	c.deferredWithdrawals.RangeOnMapping(apply)
+	return c.deferredSends.SafeSub(moduleAccount, amount)
 }
 
-func (c *ContextMemCache) ApplyOnAllDeferredOperationsAndDelete(apply func(recipient string, amount Coins)) {
-	c.RangeOnDeferredSendsAndDelete(apply)
-	c.RangeOnDeferredWithdrawalsAndDelete(apply)
+func (c *ContextMemCache) RangeOnDeferredSendsAndDelete(apply func(recipient string, amount Coins)) {
+	c.deferredSends.RangeAndRemove(apply)
```
BrandonWeng, line 76: This probably doesn't since it's supposed to be called a the end synchronously but could we add a lock or add a comment that it shouldn't be ran concurrently?

types/context_cache.go
philipsu522: shoudl we consider running go-race with this refactor?

philipsu522: we can do it when we start testing this version

types/context_cache.go
philipsu522: It should still work if called in any files ending in test.go i believe

udpatil: I can add a mutex here, but the original implementation didn't have one (but probably safer to lock)

BrandonWeng (approved):
lgtm but agree with Phil that we should run with race detector on during the testing and run a lot of bank sends/mints/burns

philipsu522 (approved):

PR #270, opened 2023-06-04
Invalid concurrent metric emission

udpatil:
This adds the invalid concurrent metric emissions

### DIFF
--- a/baseapp/baseapp.go
+++ b/baseapp/baseapp.go
@@ -997,10 +997,6 @@ func (app *BaseApp) runTx(ctx sdk.Context, mode runTxMode, txBytes []byte) (gInf
 	// Result if any single message fails or does not have a registered Handler.
 	result, err = app.runMsgs(runMsgCtx, msgs, mode)
 
-	// TODO: remove this after testing
-	tester := acltypes.Comparator{}
-	tester.EmitValidationFailMetrics()
-
 	if err == nil && mode == runTxModeDeliver {
 		// When block gas exceeds, it'll panic and won't commit the cached store.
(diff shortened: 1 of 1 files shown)


baseapp/baseapp.go
```diff
@@ -974,6 +974,7 @@ func (app *BaseApp) runTx(ctx sdk.Context, mode runTxMode, txBytes []byte) (gInf
 			if len(missingAccessOps) != 0 {
 				for op := range missingAccessOps {
 					ctx.Logger().Info((fmt.Sprintf("Antehandler Missing Access Operation:%s ", op.String())))
+					op.EmitValidationFailMetrics()
```
BrandonWeng, line 977: oh were we missing the antehandler failed metrics cus we're only emitting it here?

https://github.com/sei-protocol/sei-cosmos/blob/main/baseapp/baseapp.go#L1121

philipsu522: +1, unclear when it would/wouldn't emit now

udpatil: yeah, we were only emitting at the end of delivertx (not sure if those dependencies are included when validating so we may have been emitting anyways, let me double check)

udpatil: I dont believe it would have emitted before since it would return earlier upon error

philipsu522 (approved):

baseapp/baseapp.go
BrandonWeng: Gotcha, might be useful to add a comment but not a blocker

BrandonWeng (approved):

PR #283, opened 2023-06-12
Fix codecov

BrandonWeng:

image


codecov.yml
```diff
@@ -1,7 +1,54 @@
 coverage:
+  precision: 2
```
BrandonWeng, line 2: Looks like cosmos already havea codecov file but it was `.codecov.yml`, just going to reuse it

yzang2019 (approved):

PR #284, opened 2023-06-13
Fix allowed_msg uncapped spend limit bug

LCyson:
A grantee can use uncapped grant from a grantee when a grant is using AllowedMsgAllowance type, even with the SpendLimit specified. This issue is resulted from the types.Any Allowance object inside the AllowedMsgAllowance is not updated correctly.

This PR fixes this issue and adds according unit tests to the AllowedMsgAllowance.

- unit test
- e2e test


yzang2019 (approved):

LCyson:
> Why is this different than OSS? https://github.com/cosmos/cosmos-sdk/blob/main/x/feegrant/filtered_fee.go#L70-L87

OSS version has the same feegrant allowed_msg implementation as the one we have, so it's likely that the OSS version also has this bug

philipsu522 (approved):

LCyson:
Apply the OSS fix: https://github.com/cosmos/cosmos-sdk/pull/16097

PR #285, opened 2023-06-13
Always reset state between proposals

BrandonWeng:
Similar changes to here:
https://github.com/cosmos/cosmos-sdk/pull/15487

Our fork of cosmos-sdk has a custom integration to get ABCI++ working so I'd like @codchen to take a second look 
Unit tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,12 +37,18 @@ Ref: https://keepachangelog.com/en/1.0.0/
 
 ## [Unreleased]
 
+
+## Pending
+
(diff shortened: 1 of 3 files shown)


philipsu522 (commented):
What issues are we seeing / are we susceptible to by not having this in?

alexanderbez (approved):
Yes, this is necessary for clean state in case the proposal is rejected and needs to be created again.

codchen (approved):

yzang2019 (approved):

PR #286, opened 2023-06-13
[nice to have] multisig preventing signing with wrong key

BrandonWeng:
Doesn't seem to be absolutely necessary since it would just fail when broadcasting, more of an UI improvement. We can wait until 3.0.4 is out and then merging

Copying from: https://github.com/cosmos/cosmos-sdk/pull/12548

unit tests

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -6,7 +6,7 @@ coverage:
     project:
       default:
         threshold: 1% # allow this much decrease on project
-        target: 70%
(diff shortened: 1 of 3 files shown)


alexanderbez (approved):

philipsu522 (approved):

BrandonWeng:
Just tested this on a local node too
image
image

PR #287, opened 2023-06-14
refactor deferred balance to use memkv

udpatil:
This PR does the following:
- refactors the deferred logic to use a memKV store instead of an in memory map (allows us to avoid committing the changes when we discard other changes upon antehandler failure)
- refactors the feecollectors / deferred balance usage to key on module-txIndex-denom -> sdk.Coin 
  - when writing the deferred balances, it first adds up all of the tx specific balances, then writes them in sorted order of the module addresses
- refactors the anteDependencyGenerator to take in a txIndex to allow for generating the dependencies appropriately for feecollector resources
- adds in a txIndex to sdk.Context to contain the currently executing TX index throughout delivertx context in order to allow for writing to the partitioned KV store to avoid resource overlap
- Adds a new instantiation function for BankKeeper to instantiate with the deferredCache initialized (deferred usages panic if not set up with this new function) (have to keep old one because wasmd breaks otherwise)
- Introduces the resource types specific to the deferredCache Store Key (because this is a separate KV store from bank store it has its own resource type and isn't under the Bank resource type

Still in progress: 
- removing the context cache usages + deferred bank ops in memory structure
- adding unit tests for the memKV deferredCache that is part of bank keeper

local sei testing + adding more unit tests


baseapp/baseapp.go
```diff
@@ -970,7 +970,7 @@ func (app *BaseApp) runTx(ctx sdk.Context, mode runTxMode, txBytes []byte) (gInf
 		// Dont need to validate in checkTx mode
 		if ctx.MsgValidator() != nil && mode == runTxModeDeliver {
 			storeAccessOpEvents := msCache.GetEvents()
-			accessOps, _ := app.anteDepGenerator([]acltypes.AccessOperation{}, tx)
+			accessOps := ctx.TxMsgAccessOps()[acltypes.ANTE_MSG_INDEX]
```
udpatil, line 973: cc @BrandonWeng is there a specific reason you have previously called the dependency generator here? I assume it was because you may not have been aware that the `-1` key was reserved for the access ops generated for the ante handler ahead of time? but just want to make sure there wasnt some other reason

types/accesscontrol/comparator.go
```diff
@@ -21,6 +21,9 @@ var (
 	}
 )
 
+// use -1 to indicate that it is prior to msgs in the tx
```
alexanderbez, line 24: What is prior to messages in the transaction?

alexanderbez: Sorry, I meant the godoc should be updated to reflect that :p

baseapp/baseapp.go
BrandonWeng: Teah that was most likely the case

x/bank/keeper/deferred_cache.go
```diff
@@ -0,0 +1,126 @@
...
+
+// setBalance sets the coin balance for a module and tx Index.
+func (d *DeferredCache) setBalance(ctx sdk.Context, moduleAddr sdk.AccAddress, txIndex uint64, balance sdk.Coin) error {
+	if !balance.IsValid() {
+		return sdkerrors.Wrap(sdkerrors.ErrInvalidCoins, balance.String())
+	}
+
+	deferredStore := d.getModuleTxIndexedStore(ctx, moduleAddr, txIndex)
+	// Bank invariants require to not store zero balances, so we follow the same pattern in deferred cache.
+	if balance.IsZero() {
```
BrandonWeng, line 52: nit but could we add unit tests for failure scenarios? The happy paths seem to be covered quite well already. Not a blocker for this PR

BrandonWeng (approved):
not a blocker for this PR but could we add some unit tests for the error cases? the happy paths are tested quite well already

PR #289, opened 2023-06-14
Add more validating for staking and gov modules

BrandonWeng:
Per audit details, we're missing coverage for validation logic in staking and gov

unit tests

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -6,7 +6,7 @@ coverage:
     project:
       default:
         threshold: 1% # allow this much decrease on project
-        target: 70%
+        target: 50%
 
 comment:
   layout: "reach,diff,flags,tree,betaprofiling"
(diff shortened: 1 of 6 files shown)


types/accesscontrol/access_operation_map.go
```diff
@@ -30,18 +30,6 @@ func SendAllSignalsForTx(messageIndexToAccessOpsChannelMapping MessageAccessOpsC
 	}
 }
 
-func GetMessageAccessOps(
```
BrandonWeng, line 33: Tagging another one here, this is not used

BrandonWeng: Able to run a local chain with this removed

codecov.yml
```diff
@@ -6,7 +6,7 @@ coverage:
     project:
       default:
         threshold: 1% # allow this much decrease on project
-        target: 70%
```
BrandonWeng, line 9: Initial target was too high, the repo's only at 54%

philipsu522 (approved):

PR #294, opened 2023-06-20
Disable crisis msg verify invariant

udpatil:
This disables the handler behavior for verify invariant. It will be fairly straightforward to undo in the future if/when we would like to re-enable

localsei testing

### DIFF
--- a/x/crisis/keeper/msg_server.go
+++ b/x/crisis/keeper/msg_server.go
@@ -2,65 +2,68 @@ package keeper
 
 import (
 	"context"
+	"fmt"
 
-	sdk "github.com/cosmos/cosmos-sdk/types"
+	// sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/cosmos-sdk/x/crisis/types"
 )
 
 var _ types.MsgServer = Keeper{}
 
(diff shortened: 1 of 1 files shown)


yzang2019 (approved):

x/crisis/keeper/msg_server.go
```diff
@@ -2,65 +2,68 @@ package keeper
...
+	"fmt"
 
-	sdk "github.com/cosmos/cosmos-sdk/types"
+	// sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/cosmos-sdk/x/crisis/types"
 )
 
 var _ types.MsgServer = Keeper{}
 
 func (k Keeper) VerifyInvariant(goCtx context.Context, msg *types.MsgVerifyInvariant) (*types.MsgVerifyInvariantResponse, error) {
```
BrandonWeng, line 13: Hmm, why not disable it here? Mainly just less code to comment out 

https://github.com/sei-protocol/sei-cosmos/blob/main/x/crisis/handler.go#L12-L25

udpatil: I wanted to intentionally disable the logic that overrides the gas limit such that if someone else were to add some different execution path thats separate from the handler, theyd still need to intentionally re-enable the logic we want disabled.

PR #298, opened 2023-06-22
apply audit fixes for cache and aclmapping

udpatil:
This addresses certik audit comments about redundant logic in comparator as well as some inefficient locking in KVstore logic.

Need to test on loadtest cluster

### DIFF
--- a/store/cache/cache.go
+++ b/store/cache/cache.go
@@ -106,6 +106,8 @@ func (ckv *CommitKVStoreCache) CacheWrap(storeKey types.StoreKey) types.CacheWra
 // If the value doesn't exist in the write-through cache, the query is delegated
 // to the underlying CommitKVStore.
 func (ckv *CommitKVStoreCache) Get(key []byte) []byte {
+	ckv.mtx.Lock()
+	defer ckv.mtx.Unlock()
 
 	types.AssertValidKey(key)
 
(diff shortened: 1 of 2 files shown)


store/cache/cache.go
```diff
@@ -117,6 +115,9 @@ func (ckv *CommitKVStoreCache) Get(key []byte) []byte {
 		// cache hit
 		return value
 	}
+	// in case of cache miss, we need to lock mutex
+	ckv.mtx.Lock()
+	defer ckv.mtx.Unlock()
```
BrandonWeng, line 121: Do we not need the lock for the read on line 113?

`	value, ok := ckv.cache.Get(keyStr)`

types/accesscontrol/comparator.go
```diff
@@ -97,12 +97,6 @@ func (c *Comparator) DependencyMatch(accessOp AccessOperation, prefix []byte) bo
 		return false
 	}
 
-	// At this point the resource identifier matches, but the access type is unknown, therefore
-	// it will match both read and writes
-	if accessOp.AccessType == AccessType_UNKNOWN {
-		return true
-	}
```
BrandonWeng, line 104: qq why is this redundant?

store/cache/cache.go
udpatil: the cache is threadsafe, so it should be ok to access in this manner, I'm still thinking about the potential interactions though in the event of a write coming in at the same time that dirties the cache, so it may not be ok. This was certiks suggested fix though, but i'm still evaluating it so we shouldn't include this in the next build IMO

philipsu522 (approved):

types/accesscontrol/comparator.go
udpatil: because we return true anyways on the immediate next line

yzang2019 (approved):

BrandonWeng (approved):

PR #300, opened 2023-06-22
[Certik Audit] Fix addr len check

philipsu522:
Previously we're adding ints with bytes, but we should cast to int and add two ints.

### DIFF
--- a/x/authz/keeper/keys.go
+++ b/x/authz/keeper/keys.go
@@ -43,7 +43,7 @@ func addressesFromGrantStoreKey(key []byte) (granterAddr, granteeAddr sdk.AccAdd
 	granterAddrLen := key[1] // remove prefix key
 	kv.AssertKeyAtLeastLength(key, 3+int(granterAddrLen))
 	granterAddr = sdk.AccAddress(key[2 : 2+granterAddrLen])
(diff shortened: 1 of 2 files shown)


x/authz/keeper/keys.go
```diff
@@ -44,8 +44,8 @@ func addressesFromGrantStoreKey(key []byte) (granterAddr, granteeAddr sdk.AccAdd
 	kv.AssertKeyAtLeastLength(key, 3+int(granterAddrLen))
 	granterAddr = sdk.AccAddress(key[2 : 2+granterAddrLen])
 	granteeAddrLen := int(key[2+granterAddrLen])
-	kv.AssertKeyAtLeastLength(key, 4+int(granterAddrLen+byte(granteeAddrLen)))
-	granteeAddr = sdk.AccAddress(key[3+granterAddrLen : 3+granterAddrLen+byte(granteeAddrLen)])
+	kv.AssertKeyAtLeastLength(key, 4+int(granterAddrLen)+int(granteeAddrLen))
```
yzang2019, line 47: Do we have any test to cover this?

philipsu522: good point! added unit test

yzang2019 (approved):

PR #301, opened 2023-06-23
Make iavl separate orphan storage configurable

codchen:
We do not want to hardcode configs like orphan directory and whether separate orphan is turned on for everyone, so making it as part of app config.

A corresponding sei-chain change is needed to integrate with these config changes

local sei

### DIFF
--- a/server/config/toml.go
+++ b/server/config/toml.go
@@ -85,7 +85,13 @@ iavl-disable-fastnode = {{ .BaseConfig.IAVLDisableFastNode }}
 # Default is 0.
 compaction-interval = {{ .BaseConfig.CompactionInterval }}
 
+# deprecated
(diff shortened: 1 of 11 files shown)


server/config/toml.go
```diff
@@ -85,8 +85,14 @@ iavl-disable-fastnode = {{ .BaseConfig.IAVLDisableFastNode }}
 # Default is 0.
 compaction-interval = {{ .BaseConfig.CompactionInterval }}
 
+# deprecated
 no-versioning = {{ .BaseConfig.NoVersioning }}
 
+separate-orphan-storage = {{ .BaseConfig.SeparateOrphanStorage }}
```
yzang2019, line 91: Can we add some comments for what each config means?

yzang2019 (approved):

philipsu522 (approved):

PR #318, opened 2023-09-18
Multiversion Item Implementation and Tests

udpatil:
Add multiversion store data structures file, and implement the multiversioned item

Added unit tests to verify behavior

### DIFF
--- a/store/multiversion/data_structures.go
+++ b/store/multiversion/data_structures.go
@@ -14,7 +14,7 @@ const (
 
 type MultiVersionValue interface {
 	GetLatest() (value MultiVersionValueItem, found bool)
-	GetValueByIndex(index int) (value MultiVersionValueItem, found bool)
+	GetLatestBeforeIndex(index int) (value MultiVersionValueItem, found bool)
 	Set(index int, value []byte)
 	SetEstimate(index int)
 	Delete(index int)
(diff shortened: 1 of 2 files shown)


store/multiversion/data_structures.go
```diff
@@ -0,0 +1,110 @@
...
+	// This will ensure we get the latest indexed value relative to our current index
+	item.valueTree.DescendLessOrEqual(pivot, func(bTreeItem btree.Item) bool {
+		vItem = bTreeItem.(valueItem)
+		found = true
+		return false
+	})
+	return vItem.value, found
+}
+
+func (item *multiVersionItem) Set(index int, value []byte) {
```
codchen, line 95: can we make nextIndex a field in multiVersionItem so that the caller does not need to track it? unless there is a use case where the caller need the ability to set a prev index

udpatil: we aren't necessarily going to have each item set to an index in increasing order, right? we'll need to know when we are writing a value which TX index the value is being wrirtten for. That is what this index will represent, so the setter SHOULD have it available

philipsu522 (approved):

store/multiversion/data_structures.go
codchen: ah right that makes sense

store/multiversion/data_structures.go
```diff
@@ -0,0 +1,160 @@
...
+		return nil, false
+	}
+	valueItem := bTreeItem.(*valueItem)
+	return valueItem, true
+}
+
+// GetLatest returns the latest written value to the btree prior to the index passed in, and returns a boolean indicating whether it was found.
+//
+// A `nil` value along with `found=true` indicates a deletion that has occurred and the underlying parent store doesn't need to be hit.
+func (item *multiVersionItem) GetValueByIndex(index int) (MultiVersionValueItem, bool) {
```
codchen, line 61: nit: naming is a bit misleading here since we won't get the value at `index` even if there exists one

codchen (approved):

PR #320, opened 2023-09-28
Allow minor releases before the block height (WIP)

stevenlanders:
**DO NOT MERGE THIS YET (WIP) ** 

- If a pending minor release is running and it's before the block height, apply the upgrade
- If a pending release has been applied and before the block height, don't panic
- If a release is past the block height, regardless of type, panic 

- Unit Tests
- Local Testing

### DIFF
--- a/x/upgrade/abci.go
+++ b/x/upgrade/abci.go
@@ -23,7 +23,7 @@ import (
 func BeginBlocker(k keeper.Keeper, ctx sdk.Context, _ abci.RequestBeginBlock) {
 	defer telemetry.ModuleMeasureSince(types.ModuleName, time.Now(), telemetry.MetricKeyBeginBlocker)
 
-	plan, found := k.GetUpgradePlan(ctx)
+	plan, planFound := k.GetUpgradePlan(ctx)
 
 	if !k.DowngradeVerified() {
 		k.SetDowngradeVerified(true)
@@ -33,14 +33,14 @@ func BeginBlocker(k keeper.Keeper, ctx sdk.Context, _ abci.RequestBeginBlock) {
 		// 1. If there is no scheduled upgrade.
 		// 2. If the plan is not ready.
 		// 3. If the plan is ready and skip upgrade height is set for current height.
-		if !found || !plan.ShouldExecute(ctx) || (plan.ShouldExecute(ctx) && k.IsSkipHeight(ctx.BlockHeight())) {
+		if !planFound || !plan.MustExecute(ctx) || (plan.MustExecute(ctx) && k.IsSkipHeight(ctx.BlockHeight())) {
 			if lastAppliedPlan != "" && !k.HasHandler(lastAppliedPlan) {
 				panic(fmt.Sprintf("Wrong app version %d, upgrade handler is missing for %s upgrade plan", ctx.ConsensusParams().Version.AppVersion, lastAppliedPlan))
 			}
 		}
 	}
 
-	if !found {
+	if !planFound {
 		return
 	}
 
@@ -53,8 +53,8 @@ func BeginBlocker(k keeper.Keeper, ctx sdk.Context, _ abci.RequestBeginBlock) {
 		},
 	)
 
-	// To make sure clear upgrade is executed at the same block
-	if plan.ShouldExecute(ctx) {
+	// If the plan's block height has passed, then it must be the executed version
+	if plan.MustExecute(ctx) {
 		// If skip upgrade has been set for current height, we clear the upgrade plan
 		if k.IsSkipHeight(ctx.BlockHeight()) {
 			skipUpgradeMsg := fmt.Sprintf("UPGRADE \"%s\" SKIPPED at %d: %s", plan.Name, plan.Height, plan.Info)
@@ -65,6 +65,7 @@ func BeginBlocker(k keeper.Keeper, ctx sdk.Context, _ abci.RequestBeginBlock) {
 			return
 		}
 
+		// If we don't have an upgrade handler for this upgrade name, then we need to shutdown
 		if !k.HasHandler(plan.Name) {
 			// Write the upgrade info to disk. The UpgradeStoreLoader uses this info to perform or skip
 			// store migrations.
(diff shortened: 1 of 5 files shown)


x/upgrade/abci.go
```diff
@@ -53,49 +53,83 @@ func BeginBlocker(k keeper.Keeper, ctx sdk.Context, _ abci.RequestBeginBlock) {
...
+		applyUpgrade(k, ctx, plan)
+		return
+	}
 
+	// If running a pending minor release, apply the upgrade if handler is present
+	// Minor releases are allowed to run before the scheduled upgrade height, but not required to.
+	if plan.UpgradeDetails().IsMinorRelease() {
+		// if the pending upgrade is skipped, don't apply it
+		if k.IsSkipHeight(plan.Height) {
+			skipUpgrade(k, ctx, plan)
```
udpatil, line 77: how will this behave if the upgrade is applied 100 blocks early, but then the skip height (height of upgrade) arrives, and then the `skipUpgrade` logic runs, but the upgrade has already been applied? will this bork the binary?

x/upgrade/types/plan.go
```diff
@@ -45,3 +47,25 @@ func (p Plan) ShouldExecute(ctx sdk.Context) bool {
 func (p Plan) DueAt() string {
 	return fmt.Sprintf("height: %d", p.Height)
 }
+
+// UpgradeDetails is a struct that represents the details of an upgrade
+// This is held in the Info object of an upgrade Plan
+type UpgradeDetails struct {
```
udpatil, line 53: is upgrade info used elsewhere? is it possible theres other fields also passed in that might mess up the parsing? maybe we add a test cast like:
```
{"upgradeType": "minor", "otherInfo": "foo"}
```

stevenlanders: `encoding/json` unmarshaling is tolerant to extra & missing fields (fwiw).  The Info object is used by cosmosvisor if we choose to support auto-download (which would include a cpu-arch->binary map), but otherwise not used.

udpatil (approved):

x/upgrade/abci.go
udpatil: wait actually, question, would this cause issue with apphash because upgrading early will modify the stored state in upgrade keeper?

x/upgrade/abci.go
```diff
@@ -53,49 +53,83 @@ func BeginBlocker(k keeper.Keeper, ctx sdk.Context, _ abci.RequestBeginBlock) {
...
-			// Clear the upgrade plan at current height
-			k.ClearUpgradePlan(ctx)
+			skipUpgrade(k, ctx, plan)
 			return
 		}
+		// If we don't have an upgrade handler for this upgrade name, then we need to shutdown
+		if !k.HasHandler(plan.Name) {
+			panicUpgradeNeeded(k, ctx, plan)
+		}
+		applyUpgrade(k, ctx, plan)
```
udpatil, line 68: based on the other discussion we were doing, if the upgrade plan gets cleared after applying upgrade, wouldnt this cause issue for generating apphash since it does affect the stored state?

udpatil: My initial thought was that the upgrade action was stateless since registered plans stay registered, but if we specifically clear stuff out upon running upgrades, this may cause issue.

udpatil (changes requested):

codchen (approved):

x/upgrade/abci.go
stevenlanders: Got it - I updated with the new logic (see next commit)

x/upgrade/abci.go
```diff
@@ -53,49 +53,76 @@ func BeginBlocker(k keeper.Keeper, ctx sdk.Context, _ abci.RequestBeginBlock) {
...
-				panic(fmt.Errorf("unable to write upgrade info to filesystem: %s", err.Error()))
-			}
-
-			upgradeMsg := BuildUpgradeNeededMsg(plan)
-			// We don't have an upgrade handler for this upgrade name, meaning this software is out of date so shutdown
-			ctx.Logger().Error(upgradeMsg)
+			panicUpgradeNeeded(k, ctx, plan)
+		}
+		applyUpgrade(k, ctx, plan)
+		return
```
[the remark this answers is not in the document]
stevenlanders: We discussed in thread, but for other's purposes, we determined that this scenario and logic exists today and are electing to focus on the minimum change to allow the minor releases at the moment.

x/upgrade/types/plan.go
```diff
@@ -45,3 +47,25 @@ func (p Plan) ShouldExecute(ctx sdk.Context) bool {
...
+	UpgradeType string `json:"upgradeType"`
+}
+
+// UpgradeDetails parses and returns a details struct from the Info field of a Plan
+// The upgrade.pb.go is generated from proto, so this is separated here
+func (p Plan) UpgradeDetails() UpgradeDetails {
+	var details UpgradeDetails
+	if err := json.Unmarshal([]byte(p.Info), &details); err != nil {
+		// invalid json, assume no upgrade details
```
yzang2019, line 62: Can we add an error log here for failed to unmarshal upgrade info so that we know something bad is happening??

x/upgrade/abci.go
```diff
@@ -53,49 +53,76 @@ func BeginBlocker(k keeper.Keeper, ctx sdk.Context, _ abci.RequestBeginBlock) {
...
 	if plan.ShouldExecute(ctx) {
 		// If skip upgrade has been set for current height, we clear the upgrade plan
 		if k.IsSkipHeight(ctx.BlockHeight()) {
-			skipUpgradeMsg := fmt.Sprintf("UPGRADE \"%s\" SKIPPED at %d: %s", plan.Name, plan.Height, plan.Info)
-			ctx.Logger().Info(skipUpgradeMsg)
-
-			// Clear the upgrade plan at current height
-			k.ClearUpgradePlan(ctx)
+			skipUpgrade(k, ctx, plan)
 			return
```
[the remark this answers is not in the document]
stevenlanders: It'd be cool, but this isn't yet legal golang :)

yzang2019 (approved):

udpatil (approved):

stevenlanders:
FYI - leaving this open until integration-tests work on sei-chain

PR #322, opened 2023-10-02
[occ] Implement basic multiversion store

udpatil:
This implements the multiversion with basic functionality, but still needs additional work to implement the iterator functionality and/or persisting readsets for validation

Added unit tests for basic multiversion store

### DIFF
--- a/store/multiversion/store.go
+++ b/store/multiversion/store.go
@@ -22,6 +22,7 @@ type Store struct {
 	multiVersionMap map[string]MultiVersionValue
 	// TODO: do we need to add something here to persist readsets for later validation
 	// TODO: we need to support iterators as well similar to how cachekv does it
+	// TODO: do we need secondary indexing on index -> keys - this way if we need to abort we can replace those keys with ESTIMATE values? - maybe this just means storing writeset
 }
 
 func NewMultiVersionStore() *Store {
--- a/store/multiversion/store.go
+++ b/store/multiversion/store.go
@@ -2,8 +2,6 @@ package multiversion
 
 import (
 	"sync"
-
-	"github.com/cosmos/cosmos-sdk/internal/conv"
 )
 
 type MultiVersionStore interface {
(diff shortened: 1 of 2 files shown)


store/multiversion/store.go
```diff
@@ -0,0 +1,122 @@
...
+		s.multiVersionMap[keyString] = multiVersionValue
+	}
+}
+
+// Set implements MultiVersionStore.
+func (s *Store) Set(index int, incarnation int, key []byte, value []byte) {
+	s.mtx.Lock()
+	defer s.mtx.Unlock()
+
+	keyString := conv.UnsafeBytesToStr(key)
```
codchen, line 97: i'd prefer us to just use string casting here. If key bytes are mutated somehow the underlying map would be corrupted and would be difficult to debug

udpatil: what do you mean by mutated somehow? This is how cachekv currently implements the cache that maps between byte key and vals, so i would think there isnt too much risk in doing it this way?

udpatil: modified to use string cast for safety btw

stevenlanders (approved):
LGTM (with one question about deletion)

store/multiversion/store.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+// Delete implements MultiVersionStore.
+func (s *Store) Delete(index int, incarnation int, key []byte) {
+	s.mtx.Lock()
+	defer s.mtx.Unlock()
+
+	keyString := string(key)
+	s.tryInitMultiVersionItem(keyString)
+	s.multiVersionMap[keyString].Delete(index, incarnation)
```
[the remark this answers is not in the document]
udpatil: no, because even "deletions" are tracked explicitly, as soon as we introduce an item for a key, that key will remain in the multiversion store map for the remainder of the block, because a later transaction may need to access that data. Once all transactions have completed processing, we can discard the multiversion map after committing to store

PR #324, opened 2023-10-06
[occ] Add concurrency worker configuration

stevenlanders:
- `ConcurrencyWorkers` represents the number of workers to use for concurrent transactions
- since concurrrency-workers is a baseapp-level setting, implementations (like sei-chain) shouldn't have to pass it (but can)
- it defaults to 10 if not set (via cli default value)
- it defaults to 10 in app.toml only if that file is being created (and doesn't exist)
- if explicitly set to zero on command line, it will override with the default (for safety)
- cli takes precedence over the config file
- no one has to do anything to get it to be 10 (no config changes no sei-chain changes required (aside from new cosmos version))

- Unit Tests for setting the value
- Manually testing scenarios with sei-chain

### DIFF
--- a/baseapp/baseapp.go
+++ b/baseapp/baseapp.go
@@ -60,7 +60,8 @@ const (
 	FlagArchivalArweaveIndexDBFullPath = "archival-arweave-index-db-full-path"
 	FlagArchivalArweaveNodeURL         = "archival-arweave-node-url"
 
-	FlagChainID = "chain-id"
+	FlagChainID            = "chain-id"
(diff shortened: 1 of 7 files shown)


codchen (approved):

codchen:
Just curious how is 10 decided?

stevenlanders:
> Just curious how is 10 decided?

@codchen We're going to start benchmarking a range of values, and thought it seemed like a good starting point for that.

udpatil (approved):
lgtm

PR #328, opened 2023-10-11
[occ] Add basic worker task and scheduler shell

stevenlanders:
- Adds a basic scheduler shell (see TODOs)
- Adds a basic task definition with request/response/index
- Listens to abort channel after an execution to determine conflict

- Compiles (holding off until shape is validated)
- Basic Unit Test for ProcessAll

### DIFF
--- a/tasks/scheduler.go
+++ b/tasks/scheduler.go
@@ -35,7 +35,6 @@ type scheduler struct {
 
 // NewScheduler creates a new scheduler
 func NewScheduler(workers int, deliverTxFunc func(ctx sdk.Context, req types.RequestDeliverTx) (res types.ResponseDeliverTx)) Scheduler {
-	go func() { go func() {}() }()
 	return &scheduler{
 		workers:   workers,
 		deliverTx: deliverTxFunc,
--- a/tasks/scheduler.go
+++ b/tasks/scheduler.go
@@ -9,9 +9,17 @@ import (
 type status string
 
 const (
-	statusPending   status = "pending"
-	statusExecuted  status = "executed"
-	statusAborted   status = "aborted"
+	// statusPending tasks are ready for execution
+	// all executing tasks are in pending state
+	statusPending status = "pending"
+	// statusExecuted tasks are ready for validation
+	// these tasks did not abort during execution
+	statusExecuted status = "executed"
+	// statusAborted means the task has been aborted
+	// these tasks transition to pending upon next execution
+	statusAborted status = "aborted"
+	// statusValidated means the task has been validated
+	// tasks in this status can be reset if an earlier task fails validation
 	statusValidated status = "validated"
 )
 
@@ -91,14 +99,24 @@ func (s *scheduler) ProcessAll(ctx sdk.Context, reqs []types.RequestDeliverTx) (
 // TODO: return list of tasks that are invalid
 func (s *scheduler) validateAll(ctx sdk.Context, tasks []*deliverTxTask) ([]*deliverTxTask, error) {
 	var res []*deliverTxTask
-	for _, t := range tasks {
+
+	// find first non-validated entry
+	var startIdx int
+	for idx, t := range tasks {
+		if t.Status != statusValidated {
+			startIdx = idx
+			break
+		}
+	}
+
+	for i := startIdx; i < len(tasks); i++ {
 		// any aborted tx is known to be suspect here
-		if t.Status == statusAborted {
-			res = append(res, t)
+		if tasks[i].Status == statusAborted {
+			res = append(res, tasks[i])
 		} else {
 			//TODO: validate the tasks and add it if invalid
 			//TODO: create and handle abort for validation
-			t.Status = statusValidated
+			tasks[i].Status = statusValidated
 		}
 	}
 	return res, nil
--- a/tasks/scheduler.go
+++ b/tasks/scheduler.go
@@ -1,58 +1,31 @@
-package occ
+package tasks
 
 import (
-	"errors"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/tendermint/tendermint/abci/types"
 	"golang.org/x/sync/errgroup"
 )
 
-var (
-	ErrReadEstimate = errors.New("multiversion store value contains estimate, cannot read, aborting")
-)
-
-// Abort contains the information for a transaction's conflict
-type Abort struct {
-	DependentTxIdx int
-	Err            error
-}
-
-func NewEstimateAbort(dependentTxIdx int) Abort {
-	return Abort{
-		DependentTxIdx: dependentTxIdx,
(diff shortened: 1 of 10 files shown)


types/occ/scheduler.go
```diff
@@ -18,3 +23,147 @@ func NewEstimateAbort(dependentTxIdx int) Abort {
...
+	TaskStatusValidated TaskStatus = "validated"
+)
+
+type Task struct {
+	Status      TaskStatus
+	Index       int
+	Incarnation int
+	Request     types.RequestDeliverTx
+	Response    *types.ResponseDeliverTx
+}
```
udpatil, line 42: do we need to store any data here about "dependent TX" for the ESTIMATE abort case?

types/occ/scheduler.go
```diff
@@ -18,3 +23,147 @@ func NewEstimateAbort(dependentTxIdx int) Abort {
...
+	}
+
+	for i := 0; i < workers; i++ {
+		grp.Go(func() error {
+			for {
+				select {
+				case <-gCtx.Done():
+					return gCtx.Err()
+				case task, ok := <-ch:
+					if !ok {
```
udpatil, line 134: can we refactor this logic under `case` to a function like `receiveTask`?

types/occ/scheduler.go
```diff
@@ -18,3 +23,147 @@ func NewEstimateAbort(dependentTxIdx int) Abort {
...
+	return res, nil
+}
+
+// ExecuteAll executes all tasks concurrently
+// Tasks are updated with their status
+// TODO: retries on aborted tasks
+// TODO: error scenarios
+func (s *scheduler) executeAll(ctx sdk.Context, tasks []*Task) error {
+	ch := make(chan *Task, len(tasks))
+	grp, gCtx := errgroup.WithContext(ctx.Context())
```
udpatil, line 119: shouldn't we create the context `WithCancel` so that we can kill off the workers from the scheduler? we check the context.Done case below, but we would still need to indicate `Done` for the context via Cancel right?

types/occ/scheduler.go
```diff
@@ -18,3 +23,147 @@ func NewEstimateAbort(dependentTxIdx int) Abort {
...
+			for {
+				select {
+				case <-gCtx.Done():
+					return gCtx.Err()
+				case task, ok := <-ch:
+					if !ok {
+						return nil
+					}
+					//TODO: ensure version multi store is on context
+					//abortCh := make(chan Abort)
```
udpatil, line 138: we can wrap mvkv here prior to calling the delivertx

types/occ/scheduler.go
```diff
@@ -18,3 +23,147 @@ func NewEstimateAbort(dependentTxIdx int) Abort {
...
+					resp := s.deliverTx(ctx, task.Request)
```
udpatil, line 139: wouldnt it be better here to call `deliverTx` async, otherwise there would be a deadlock since we'd never get to the channel read below? and then below we have a select statement with abortChannel, deliverTx response, or context Canceled?

types/occ/scheduler.go
```diff
@@ -18,3 +23,147 @@ func NewEstimateAbort(dependentTxIdx int) Abort {
...
+
+		// execute sets statuses of tasks to either executed or aborted
+		err := s.executeAll(ctx, toExecute)
+		if err != nil {
+			return err
+		}
+
+		// validate returns any that should be re-executed
+		// note this processes ALL tasks, not just those recently executed
+		toExecute, err = s.validateAll(ctx, tasks)
```
udpatil, line 83: we want validation to also be ongoing and not done after all executions are done, that way if we fail validation for tx X we can immediately re-execute instead of having waited for ALL the txs to have executed before realizing the validation failed

types/occ/scheduler.go
stevenlanders: yes once we do the estimate/validation pieces there are a few things we'll need to track

types/occ/scheduler.go
stevenlanders: With errgroups, one can just return an error like context timeout error, and all ctx.Done()s will cancel.  If something outside a task needs to cancel, then yes we can wrap errgroup with a WithCancel context, but then it's still going to be the ctx.Done() that exits those routines.

types/occ/scheduler.go
stevenlanders: Is it clear that as soon as validation fails is when we know it should be immediately re-executed?  (If we have to prioritize re-executions?)  That wasn't yet clear to me because it seems like we have to do a full-validation at the end anyway.

tasks/scheduler.go
```diff
@@ -1,58 +1,39 @@
...
+type status string
 
 const (
-	TaskStatusPending   TaskStatus = "pending"
-	TaskStatusExecuted  TaskStatus = "executed"
-	TaskStatusAborted   TaskStatus = "aborted"
-	TaskStatusValidated TaskStatus = "validated"
+	// statusPending tasks are ready for execution
+	// all executing tasks are in pending state
+	statusPending status = "pending"
```
udpatil, line 14: shouldnt these be `StatusPending` so theyre public?

tasks/scheduler.go
```diff
@@ -68,43 +49,75 @@ func NewScheduler(workers int, deliverTxFunc func(ctx sdk.Context, req types.Req
...
+	return collectResponses(tasks), nil
 }
 
-// TODO: validate each task
+// TODO: validate each tasks
 // TODO: return list of tasks that are invalid
-func (s *scheduler) validateAll(ctx sdk.Context, tasks []*Task) ([]*Task, error) {
-	var res []*Task
-	for _, t := range tasks {
+func (s *scheduler) validateAll(ctx sdk.Context, tasks []*deliverTxTask) ([]*deliverTxTask, error) {
```
udpatil, line 101: this isnt actually doing any validation, should we rename this to something like collectTasksForValidation (since thats what it appears to be doing?)

types/occ/types.go
```diff
@@ -0,0 +1,22 @@
...
+import (
+	"errors"
+)
+
+var (
+	ErrReadEstimate = errors.New("multiversion store value contains estimate, cannot read, aborting")
+)
+
+// Abort contains the information for a transaction's conflict
+type Abort struct {
```
udpatil, line 12: i'm ok moving this to scheduler too if we want, otherwise if theres a concern of import cycle we can keep in types so it can be used by both `store` and `scheduler`

tasks/scheduler.go
stevenlanders: it's going to be doing the validation, it's collecting tasks for re-execution

tasks/scheduler.go
stevenlanders: at the moment tasks don't leak the scheduler so nothing else cares

udpatil (approved):

PR #329, opened 2023-10-12
[occ] Implement iterator for mvkv

udpatil:
This implements Iterator and ReverseIterator for mvkv for the KVStore interface. The memiterator will be composed of versionindexedstore and multiversionstore, and will yield values in a cascading fashion firstly from the writeset, and then second from the multiversion store.

This still needs optimization to persisted sorted keys instead of reconstructing sorted keys each time.

Unit test to verify basic functionality

### DIFF
--- a/store/multiversion/memiterator.go
+++ b/store/multiversion/memiterator.go
@@ -1,0 +1,74 @@
+package multiversion
+
+import (
+	dbm "github.com/tendermint/tm-db"
+
+	"github.com/cosmos/cosmos-sdk/store/types"
+	scheduler "github.com/cosmos/cosmos-sdk/types/occ"
+)
+
+// Iterates over iterKVCache items.
+// if key is nil, means it was deleted.
+// Implements Iterator.
+type memIterator struct {
+	types.Iterator
+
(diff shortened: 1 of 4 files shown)


store/multiversion/memiterator.go
```diff
@@ -0,0 +1,74 @@
...
+	if val, ok := mi.writeset[string(key)]; ok {
+		return val
+	}
+
+	// get the value from the multiversion store
+	val := mi.mvStore.GetLatestBeforeIndex(mi.index, key)
+
+	// if we have an estiamte, write to abort channel
+	if val.IsEstimate() {
+		mi.abortChannel <- scheduler.NewEstimateAbort(val.Index())
```
stevenlanders, line 66: should this keep processing here?  (or maybe something upstream handles?)

udpatil: we can't because once we have an estimate, we can't know what the actual value will be that we need to read, so this abort would be expected to be handled within the task / scheduler level, and this tx index would need to wait for the "estimate" to be cleared

PR #331, opened 2023-10-12
Fast reject invalid consensus params

philipsu522:
Previously, a  proposal to change Tendermint consensus params to an invalid value could be made, but could potentially halt the chain b/c at the block in which the proposal passes, Tendermint will verify that the params are valid and panic if the change is not valid (see https://github.com/sei-protocol/sei-tendermint/blob/d426f1fe475eb0c406296770ff5e9f8869b3887e/internal/state/execution.go#L320), thus halting the change. This PR early rejects the proposal so that the chain will no longer halt.
- Added unit tests
- Verified on a internal chain:
```
root@ip-172-31-34-209:/home/ubuntu/sei-chain# seid tx gov submit-proposal param-change prop.json --from admin -b block --fees 20000usei
Enter keyring passphrase:
Error: block.MaxGas must be greater or equal to -1. Got -2
...
root@ip-172-31-34-209:/home/ubuntu/sei-chain# seid tx gov submit-proposal param-change prop.json --from admin -b block --fees 2000usei
Enter keyring passphrase:
Error: block.MaxBytes is too big. 2202009600000 > 104857600
```

### DIFF
--- a/x/params/types/proposal/proposal.go
+++ b/x/params/types/proposal/proposal.go
@@ -1,7 +1,9 @@
 package proposal
 
 import (
+	"encoding/json"
 	"fmt"
+	"github.com/tendermint/tendermint/types"
 	"strings"
 
 	yaml "gopkg.in/yaml.v2"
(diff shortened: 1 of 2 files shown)


yzang2019 (approved):

stevenlanders (approved):
lgtm

PR #332, opened 2023-10-16
[occ] Add validation logic to scheduler

stevenlanders:
## Describe your changes and provide context
- This adds a versioned store for each incarnation of a transaction
- Scheduler re-executes based on validation output

## Testing performed to validate your change
- Unit test for scheduler (confirms expected tx output for conflicting transactions)

PR #336, opened 2023-10-18
[occ] Add scheduler logic for validation

stevenlanders:
- This was copied from #332 which became unwieldy due to commit history (merges/rebases)
- Adds scheduler logic for validation
- In this initial version it completes all executions then performs validations (which feed retries)
  -  Once we start benchmarking we can make performance improvements to this 
- Retries tasks that fail validation and have no dependencies

- Scheduler Test verifies multi-worker with conflicts

### DIFF
--- a/server/mock/store.go
+++ b/server/mock/store.go
@@ -229,3 +229,13 @@ func (kv kvStore) ReverseSubspaceIterator(prefix []byte) sdk.Iterator {
 func NewCommitMultiStore() sdk.CommitMultiStore {
 	return multiStore{kv: make(map[sdk.StoreKey]kvStore)}
 }
+
+func (ms multiStore) SetKVStores(handler func(key store.StoreKey, s sdk.KVStore) store.CacheWrap) store.MultiStore {
+	//TODO implement me
+	panic("implement me")
+}
+
+func (ms multiStore) StoreKeys() []sdk.StoreKey {
+	//TODO implement me
+	panic("implement me")
+}
--- a/server/mock/store.go
+++ b/server/mock/store.go
@@ -231,11 +231,9 @@ func NewCommitMultiStore() sdk.CommitMultiStore {
 }
 
 func (ms multiStore) SetKVStores(handler func(key store.StoreKey, s sdk.KVStore) store.CacheWrap) store.MultiStore {
-	//TODO implement me
-	panic("implement me")
+	panic("not implemented")
 }
 
 func (ms multiStore) StoreKeys() []sdk.StoreKey {
-	//TODO implement me
-	panic("implement me")
+	panic("not implemented")
 }
--- a/store/multiversion/mvkv.go
+++ b/store/multiversion/mvkv.go
@@ -1,11 +1,12 @@
 package multiversion
 
 import (
-	abci "github.com/tendermint/tendermint/abci/types"
 	"io"
 	"sort"
 	"sync"
 	"time"
+
+	abci "github.com/tendermint/tendermint/abci/types"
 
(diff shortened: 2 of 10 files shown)


tasks/scheduler.go
```diff
@@ -49,6 +74,28 @@ func NewScheduler(workers int, deliverTxFunc func(ctx sdk.Context, req types.Req
...
+		mv.InvalidateWriteset(task.Index, task.Incarnation)
+	}
+}
+
+func (s *scheduler) findConflicts(task *deliverTxTask) []int {
+	var conflicts []int
+	uniq := make(map[int]struct{})
+	for _, mv := range s.multiVersionStores {
+		mvConflicts := mv.ValidateTransactionState(task.Index)
+		for _, c := range mvConflicts {
```
udpatil, line 88: note: conflicts are already deduped + sorted, so this is unnecessary

server/mock/store.go
```diff
@@ -229,3 +229,13 @@ func (kv kvStore) ReverseSubspaceIterator(prefix []byte) sdk.Iterator {
 func NewCommitMultiStore() sdk.CommitMultiStore {
 	return multiStore{kv: make(map[sdk.StoreKey]kvStore)}
 }
+
+func (ms multiStore) SetKVStores(handler func(key store.StoreKey, s sdk.KVStore) store.CacheWrap) store.MultiStore {
+	//TODO implement me
```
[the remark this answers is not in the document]
stevenlanders: these were generated, no need to implement (just removed todo and renamed 'not implemented')

tasks/scheduler.go
stevenlanders: it's necessary to dedupe/sort across storekeys

store/multiversion/mvkv.go
```diff
@@ -359,18 +360,20 @@ func (store *VersionIndexedStore) UpdateReadSet(key []byte, value []byte) {
...
 }
 
 // ResetEvents implements types.CacheWrap so this store can exist on the cache multi store
 func (store *VersionIndexedStore) ResetEvents() {
-	//TODO implement me
-	panic("implement me")
+	panic("not implemented")
+}
+
+func (store *VersionIndexedStore) UpdateIterateSet(iterationTracker iterationTracker) {
```
udpatil, line 376: wasnt this already present from occ-main, or is the diff highlight just weird because of merge conflict resolution

tasks/scheduler.go
```diff
@@ -193,28 +197,24 @@ func (s *scheduler) validateAll(tasks []*deliverTxTask) ([]*deliverTxTask, error
...
 					// otherwise, wait for completion
-					tasks[i].Conflicts = conflicts
+					tasks[i].Dependencies = conflicts
 					tasks[i].Status = statusWaiting
 				}
-			} else {
-				// no conflicts means this task is validated
+			} else if len(conflicts) == 0 {
 				tasks[i].Status = statusValidated
 			}
```
udpatil, line 213: in the case where we are valid but have conflicts, does the task states stay as `executed` then? (just want to make sure my understanding is correct

udpatil (approved):

tasks/scheduler.go
stevenlanders: that's right, it remains as executed and not returned for re-execution, and therefore is eligible for re-validation

store/multiversion/mvkv.go
stevenlanders: This line doesn't seem to show up in the overall diff, just in the commit itself (a merge from occ-main).

PR #337, opened 2023-10-19
[occ] Iterateset tracking and validation implementation

udpatil:
This implements a tracked iterator that is used to keep track of keys that have been iterated, and to also save metadata about the iteration for LATER validation. The iterator will be replayed and if there are any new keys / any keys missing within the iteration range, it will fail validation. the actual values served by the iterator are covered by readset validation. 

Additionally, the early stop behavior allows the iterateset to ONLY be sensitive to changes to the keys available WITHIN the iteration range. In the event that we perform iteration, and THEN write a key within the range of iteration, this will not fail iteration because we take a snapshot of the mvkv writeset at the moment of iteration, so when we replay the iterator, we populate that iterator with the writeset at that time, so we appropriately replicate the iterator behavior.

In the case that we encounter an ESTIMATE, we have to terminate the iterator validation and mark it as failed because it is impossible to know whether that ESTIMATE represents a value change or a delete, since the latter, will affect the keys available for iteration.

This change also implements handlers that iterators receive for updating readset and iterateset in the `mvkv`

Unit tests for various iteration scenarios

### DIFF
--- a/store/multiversion/memiterator.go
+++ b/store/multiversion/memiterator.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/cosmos/cosmos-sdk/store/types"
 	occtypes "github.com/cosmos/cosmos-sdk/types/occ"
+	scheduler "github.com/cosmos/cosmos-sdk/types/occ"
 )
 
 // Iterates over iterKVCache items.
@@ -14,15 +15,17 @@ type memIterator struct {
 	types.Iterator
 
 	mvStore      MultiVersionStore
-	writeset     map[string][]byte
+	writeset     WriteSet
 	index        int
 	abortChannel chan occtypes.Abort
+	ReadsetHandler
 }
 
 func (store *VersionIndexedStore) newMemIterator(
 	start, end []byte,
 	items *dbm.MemDB,
 	ascending bool,
+	readsetHandler ReadsetHandler,
 ) *memIterator {
 	var iter types.Iterator
 	var err error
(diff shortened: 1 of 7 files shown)


store/multiversion/memiterator.go
```diff
@@ -66,9 +69,47 @@ func (mi *memIterator) Value() []byte {
 		mi.abortChannel <- occtypes.NewEstimateAbort(val.Index())
 	}
 
+	// need to update readset
 	// if we have a deleted value, return nil
 	if val.IsDeleted() {
+		mi.ReadsetHandler.UpdateReadSet(key, nil)
```
codchen, line 75: nit can this be a defer?

store/multiversion/store.go
```diff
@@ -212,9 +219,109 @@ func (s *Store) GetReadset(index int) ReadSet {
...
+	return s.txIterateSets[index]
+}
+
+// CollectIteratorItems implements MultiVersionStore. It will return a memDB containing all of the keys present in the multiversion store within the iteration range prior to (exclusive of) the index.
+func (s *Store) CollectIteratorItems(index int) *db.MemDB {
+	sortedItems := db.NewMemDB()
+
+	// get all writeset keys prior to index
+	keys := s.GetAllWritesetKeys()
+	for i := 0; i < index; i++ {
```
codchen, line 242: should we also check `i < len(keys)`?

codchen (approved):
overall looks good. Left some nits

store/multiversion/store.go
udpatil: why is that? `i` is used to access the writeset corresponding to a specific transaction index in the multiversion store, theres no guarantee that all indices i < len(keys) would be present this moment, in which case we should skip to the next index. I'll add in an `ok` check for the map value before iterating over the indexedWriteset for explicit checking of presence, but it should still no-op given that a range over a nil slice would no-op, right?

store/multiversion/memiterator.go
udpatil: yup, changed, although i'm not certain that it changes the outcome that much?

stevenlanders (approved):

PR #339, opened 2023-10-20
[occ] Allow parent app to pass per-tx contexts

stevenlanders:
- sei-chain needs to pass per-tx notification channels on contexts 
- this allows for per-tx contexts to be passed
 
- updated all unit tests


stevenlanders:
we're going to hold off on this, until there's a per-tx context needed (since we don't need the Dag channels)

PR #341, opened 2023-10-24
[occ] Add struct field and helpers for estimate prefills

udpatil:
This adds in the ability to prefill estimates based on metadata passed along with deliverTxBatch

Unit Test to verify that multiversion store initialization is now idempotent, and works properly regardless of whether estimate prefill is enabled


types/tx_batch.go
```diff
@@ -1,11 +1,21 @@
...
+	"github.com/cosmos/cosmos-sdk/store/multiversion"
+	abci "github.com/tendermint/tendermint/abci/types"
+)
 
 // DeliverTxEntry represents an individual transaction's request within a batch.
 // This can be extended to include tx-level tracing or metadata
 type DeliverTxEntry struct {
-	Request abci.RequestDeliverTx
+	Request  abci.RequestDeliverTx
+	Metadata DeliverTxMetadata
```
stevenlanders, line 12: 2 cents: I'd probably would have just put the Writeset here, because Metadata normally means things like headers/trackers/tracers and not as much real-arguments to the function.

tasks/scheduler.go
```diff
@@ -57,6 +57,7 @@ func (dt *deliverTxTask) Increment() {
 
 // Scheduler processes tasks concurrently
 type Scheduler interface {
+	PrefillEstimates(ctx sdk.Context, metadatas []sdk.DeliverTxMetadata)
```
stevenlanders, line 60: 2 cents: if this always must happen prior to ProcessAll, it may be worth just including the logic in ProcessAll to simplify the interaction.

stevenlanders (approved):
Assuming -1 for incarnation is okay, this looks good to me.

PR #343, opened 2023-10-30
Fix map access panic

stevenlanders:
- `CollectIteratorItems` needs to hold an RLock to avoid a concurrent access panic 

- Reproduced through a sei-chain-side test (concurrent instantiates)

### DIFF
--- a/store/multiversion/store.go
+++ b/store/multiversion/store.go
@@ -319,21 +319,27 @@ func (s *Store) validateIterator(index int, tracker iterationTracker) bool {
 	}
 }
 
-// TODO: do we want to return bool + []int where bool indicates whether it was valid and then []int indicates only ones for which we need to wait due to estimates? - yes i think so?
-func (s *Store) ValidateTransactionState(index int) (bool, []int) {
-	defer telemetry.MeasureSince(time.Now(), "store", "mvs", "validate")
-	conflictSet := map[int]struct{}{}
-	valid := true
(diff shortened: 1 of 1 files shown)


store/multiversion/store.go
```diff
@@ -243,10 +243,13 @@ func (s *Store) GetIterateset(index int) Iterateset {
...
 // CollectIteratorItems implements MultiVersionStore. It will return a memDB containing all of the keys present in the multiversion store within the iteration range prior to (exclusive of) the index.
 func (s *Store) CollectIteratorItems(index int) *db.MemDB {
+	s.mtx.RLock()
+	defer s.mtx.RUnlock()
+
 	sortedItems := db.NewMemDB()
 
 	// get all writeset keys prior to index
-	keys := s.GetAllWritesetKeys()
+	keys := s.txWritesetKeys
```
udpatil, line 252: this will acquire the lock for longer right? whereas none of the below logic actually invokes resources protected by the lock? does it make sense to do this way?

udpatil: meant to point to lines 246/247 where lock is acquired

stevenlanders: In Go, maps are always returned by reference, not by value.  This means returning a reference via GetAllWritesetKeys doesn't give us a copy, and doesn't do anything to protect the map.  The concurrent-access here is the `keys[i]` moment.

PR #348, opened 2023-11-10
[occ] Fix parent store readset validation

udpatil:
This fixes the validation to remove a panic for a case that can actually occur if a transaction writes a key that is later read, and that writing transaction is reverted and then the readset validation reads from parent store. In this case, the readset would have a conflict based on the data available in parent store, so we shouldn't panic. This also adds in the resource types needed for the new DEX_MEM keys

Tested in loadtest cluster


stevenlanders (approved):
lgtm (I would just suggest a unit test for the scenario if possible)

udpatil:
> lgtm (I would just suggest a unit test for the scenario if possible)

yup good point, will add

PR #360, opened 2023-11-22
change DeliverTx to take typed tx

codchen:
- To avoid multiple deserialization, as well as adding cached values to sdk.Tx
- Add new acl constants for evm subprefixes
- Add a new bank send method that doesn't automatically create accounts
unit tests & local sei integration


stevenlanders:
(@udpatil note this probably will affect OCC's deliverTxBatch interface when we eventually merge these together)

baseapp/baseapp.go
```diff
@@ -942,7 +939,7 @@ func (app *BaseApp) runTx(ctx sdk.Context, mode runTxMode, txBytes []byte) (gInf
 		// NOTE: Alternatively, we could require that AnteHandler ensures that
 		// writes do not happen if aborted/failed.  This may have some
 		// performance benefits, but it'll be more difficult to get right.
-		anteCtx, msCache = app.cacheTxContext(ctx, txBytes)
+		anteCtx, msCache = app.cacheTxContext(ctx, checksum)
```
jewei1997, line 942: Can you explain why we use the SHA256 over the txBytes?

codchen: @jewei1997 in `cacheTxContext`, `txBytes` is only used to calculate its sha256, so we just pass in the calculated sha256 directly

PR #379, opened 2023-12-12
[OCC] Fix hang where abort channel blocks iterator

stevenlanders:
- instead of assuming one thing will arrive to the abort channel, drain it
- new unit test captures situation (tests iterator)

### DIFF
--- a/tasks/scheduler.go
+++ b/tasks/scheduler.go
@@ -485,17 +485,17 @@ func (s *scheduler) executeTask(task *deliverTxTask) {
 	var abortOccurred bool
 
 	var wg sync.WaitGroup
+	wg.Add(1)
+
+	var abort *occ.Abort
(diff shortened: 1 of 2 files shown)


tasks/scheduler.go
```diff
@@ -442,7 +441,8 @@ func (s *scheduler) prepareTask(task *deliverTxTask) {
 }
 
 // executeTask executes a single task
-func (s *scheduler) executeTask(task *deliverTxTask) {
+func (s *scheduler) executeTask3(task *deliverTxTask) {
```
udpatil, line 444: is this just for preserving the old impl?

udpatil (approved):

tasks/scheduler.go
stevenlanders: nope- this was an oversight that needs removing :)

PR #388, opened 2023-12-28
Improve cachekv write performance

yzang2019:
This PR avoid the several O(N) operations to clear up the maps. Instead of looping through each item and delete them, it will be much more efficient to just reset to a new map.

Tested in loadtest cluster and get performance improved (3ms -> 0.1ms)

### DIFF
--- a/store/cachekv/store.go
+++ b/store/cachekv/store.go
@@ -68,7 +68,7 @@ func (store *Store) GetStoreType() types.StoreType {
 func (store *Store) getFromCache(key []byte) []byte {
 	if cv, ok := store.cache.Load(conv.UnsafeBytesToStr(key)); ok {
 		return cv.(*types.CValue).Value()
-	} 
+	}
 	return store.parent.Get(key)
 }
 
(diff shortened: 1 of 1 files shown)


philipsu522 (approved):

store/cachekv/store.go
```diff
@@ -135,18 +135,9 @@ func (store *Store) Write() {
 	// Clear the cache using the map clearing idiom
 	// and not allocating fresh objects.
 	// Please see https://bencher.orijtech.com/perfclinic/mapclearing/
-	store.cache.Range(func(key, value any) bool {
```
udpatil, line 138: do these maps no longer behave in the idiomatic way? I gues that makes sense as we migrated from native go maps to sync maps, but just want to confirm

udpatil (approved):

store/cachekv/store.go
```diff
@@ -135,18 +135,9 @@ func (store *Store) Write() {
 	// Clear the cache using the map clearing idiom
```
stevenlanders, line 135: (i think this comment can probably go away now - LGTM :+1:)

stevenlanders (approved):

store/cachekv/store.go
yzang2019: Yeah, there's also a mutex to protect the writes

PR #391, opened 2024-01-03
Fix memory leak due to emitting infinite events during snapshot

yzang2019:
Problem:
Currently when the wasmd snapshotter export snapshot, it will create a CacheMultiStore and use cachekv to access the historical state, which will then create a mergeIterator to iterate through key/values. In MergeIterator, everytime we access a key or value, it will emit a resource access events for "iterate" operation, but there's no limit on how many events to be emitted, so as we append more and more events to event manager during wasm snapshot, it starts to fill up the memory.

Solution:
We can completely get rid of this iterator events, that event is not useful at all. 

Tested and validated on atlantic-2


store/cachekv/mergeiterator.go
```diff
@@ -138,14 +138,12 @@ func (iter *cacheMergeIterator) Value() []byte {
 	// If parent is invalid, get the cache value.
 	if !iter.parent.Valid() {
 		value := iter.cache.Value()
-		iter.eventManager.EmitResourceAccessReadEvent("iterator", iter.storeKey, iter.cache.Key(), value)
```
yzang2019, line 141: These events are causing memory leak during snapshot creation, it should be safe to delete since those events are not being used in critical path at all

codchen (approved):

PR #394, opened 2024-01-11
Add migration handler for disabling seqno

philipsu522:
Add migration handler to add new disable seqno parameter to chain. Note that i verified that we already use default values so we can just set the params to default instead of maintaining authParamsV1 and V2
- unit tests
- upgraded local chain, ensure that bank sends work

### DIFF
--- a/x/auth/keeper/v2_to_v3_test.go
+++ b/x/auth/keeper/v2_to_v3_test.go
@@ -1,0 +1,28 @@
+package keeper_test
+
+import (
+	"github.com/cosmos/cosmos-sdk/x/auth/keeper"
+	"github.com/cosmos/cosmos-sdk/x/auth/types"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestMigrate2to3(t *testing.T) {
+	app, ctx := createTestApp(true)
+
+	prevParams := types.Params{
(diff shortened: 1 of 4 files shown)


x/params/types/subspace.go
```diff
@@ -219,6 +219,7 @@ func (s Subspace) Update(ctx sdk.Context, key, value []byte) error {
 // in the ParamSetPair by calling Subspace#Get.
 func (s Subspace) GetParamSet(ctx sdk.Context, ps ParamSet) {
 	for _, pair := range ps.ParamSetPairs() {
+		fmt.Printf("PSUDEBUG - pairs: %s\n", pair)
```
yzang2019, line 222: Shall we remove this?

x/auth/keeper/v2_to_v3_test.go
```diff
@@ -0,0 +1,28 @@
+package keeper_test
+
+import (
+	"github.com/cosmos/cosmos-sdk/x/auth/keeper"
+	"github.com/cosmos/cosmos-sdk/x/auth/types"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestMigrate2to3(t *testing.T) {
```
yzang2019, line 10: nice!

yzang2019 (approved):

PR #398, opened 2024-01-18
Occ enabled cfg

udpatil:
This adds `occ-enabled` as a config for baseapp to control whether to execute transactions with OCC parallelism.

Tested on sei-chain


stevenlanders:
Does setting to false work?  (does it see that as an explicit override to false or does  the default of `true` win because false is a default bool value and looks unset?)

udpatil:
yeah, it reads a false value specifically if we change the config value to false

stevenlanders (approved):

PR #401, opened 2024-01-23
fix(baseapp): Ensure Panic Recovery in Prepare & Process Handlers

alexanderbez:
## Changelog

* Ensure we have panic recovery handlers in `PrepareProposal` and `ProcessProposal`.


yzang2019 (approved):
LGTM

PR #443, opened 2024-02-27
Amplify cosmos tx priority

codchen:
Since EVM transactions specify gas price in unit of wei, we have started representing priority of EVM transactions in wei-per-gas. In order for cosmos transactions' priority to be comparable with EVM transactions', we would like to use the same unit for priority here as well. Hence in this PR we amplify cosmos transaction priority by 10^12 (if the original priority is based on the base denom aka usei)

unit test

### DIFF
--- a/x/auth/ante/validator_tx_fee.go
+++ b/x/auth/ante/validator_tx_fee.go
@@ -8,6 +8,8 @@ import (
 	paramskeeper "github.com/cosmos/cosmos-sdk/x/params/keeper"
 )
 
+var BaseDenomGasPriceAmplfier = sdk.NewInt(1_000_000_000_000)
+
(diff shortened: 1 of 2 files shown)


jewei1997 (approved):
LGTM!

jewei1997:
Just FYI seeing this on a cosmos bank send with this cmd:
```
seid tx bank send $(seid keys show admin -a) sei1m9qugvk4h66p6hunfajfg96ysc48zeq4m0d82c 1usei --fees 20000usei -b block -y
```

this is my output:
```
c.Amount is 20000 BaseDenomGasPriceAmplfier is 1000000000000 gas is 200000
gasPrice is 100000000000 <- 10^11
priority is 100000000000 <- 10^11
```

PR #452, opened 2024-03-05
[OCC] Add sync fallback for high incarnation counts

stevenlanders:
- If incarnation exceeds a threshold, revert to synchronous as a safety net
- Appends dependencies to list rather than replace to avoid forgetting old deps

- High conflict test no longer takes a longer time to run

### DIFF
--- a/tasks/scheduler.go
+++ b/tasks/scheduler.go
@@ -287,8 +287,8 @@ func (s *scheduler) ProcessAll(ctx sdk.Context, reqs []*sdk.DeliverTxEntry) ([]t
 
 	toExecute := tasks
 	for !allValidated(tasks) {
-		// if the max incarnation > 5, we should revert to synchronous
-		if s.maxIncarnation > maximumIncarnation {
+		// if the max incarnation >= 5, we should revert to synchronous
+		if s.maxIncarnation >= maximumIncarnation {
 			// process synchronously
 			s.synchronous = true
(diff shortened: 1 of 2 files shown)


tasks/scheduler.go
```diff
@@ -193,15 +206,25 @@ func (s *scheduler) tryInitMultiVersionStore(ctx sdk.Context) {
 	s.multiVersionStores = mvs
 }
 
-func indexesValidated(tasks []*deliverTxTask, idx []int) bool {
-	for _, i := range idx {
+func dependenciesValidated(tasks []*deliverTxTask, deps map[int]struct{}) bool {
+	for i := range deps {
```
udpatil, line 210: does the order of this validation matter? map iteration isn't guaranteed in order so if it does we should make a slice + sort first

tasks/scheduler_test.go
```diff
@@ -176,7 +168,7 @@ func TestProcessAll(t *testing.T) {
 		{
 			name:      "Test every tx accesses same key",
 			workers:   50,
-			runs:      1,
+			runs:      5,
```
udpatil, line 171: did we also want to make an assertion for this test that maxIteration is like 6 or something?

tasks/scheduler.go
stevenlanders: it does not matter because all must be validated to move forward

udpatil (approved):

PR #454, opened 2024-03-08
Fix indexesValidated and PrefillEstimates to operate on absolute idx

udpatil:
This is one component that was missed when refactoring to use absolute indices for EVM changes. This change refactors such that prefill estimates will appropriately fill the estimates by absolute Index and indexes validated will similarly check via absolute indices instead of relative.

Existing unit tests + loadtesting

### DIFF
--- a/tasks/scheduler.go
+++ b/tasks/scheduler.go
@@ -92,6 +92,7 @@ type scheduler struct {
 	workers            int
 	multiVersionStores map[sdk.StoreKey]multiversion.MultiVersionStore
 	tracingInfo        *tracing.Info
+	allTasksMap        map[int]*deliverTxTask
 	allTasks           []*deliverTxTask
 	executeCh          chan func()
 	validateCh         chan func()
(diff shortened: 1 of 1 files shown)


tasks/scheduler.go
```diff
@@ -219,12 +225,12 @@ func allValidated(tasks []*deliverTxTask) bool {
...
 func (s *scheduler) PrefillEstimates(reqs []*sdk.DeliverTxEntry) {
 	// iterate over TXs, update estimated writesets where applicable
-	for i, req := range reqs {
+	for _, req := range reqs {
 		mappedWritesets := req.EstimatedWritesets
 		// order shouldnt matter for storeKeys because each storeKey partitioned MVS is independent
 		for storeKey, writeset := range mappedWritesets {
 			// we use `-1` to indicate a prefill incarnation
-			s.multiVersionStores[storeKey].SetEstimatedWriteset(i, -1, writeset)
+			s.multiVersionStores[storeKey].SetEstimatedWriteset(req.AbsoluteIndex, -1, writeset)
```
philipsu522, line 233: would it be easy to add a unit test for this?

philipsu522 (approved):

tasks/scheduler.go
udpatil: yeah, good point, we have a simple unit test to assert the proper values in the MultiVersionStore, I'll add one in

yzang2019 (approved):

stevenlanders (approved):

PR #464, opened 2024-03-19
Fix DeleteAll for CacheKV/MultiVersionKV so that all keys are properly deleted

codchen:
If we have a cacheKV that has a dirty cache of 1->a and a parent with an entry 2->b, the previous implementation of `DeleteAll` would only call cacheKV.Delete on key 1 but not affecting 2 at all (i.e. if someone queries key 2 on the cacheKV they would still get `b`, but the expectation is for 2 to be deleted as appearing on the cacheKV level). This PR fixes that by first getting all the keys to be (marked as) deleted recursively and then delete all those keys one-by-one; note that this recursion should be fairly efficient since it's not subject to the exponential latency that merge iterators have.

This PR also fixed DeleteAll for `mvkv` so that it doesn't actually touch parent store's value but only change the writeset (as a single `Delete` would have done).

unit test

### DIFF
--- a/store/multiversion/mvkv.go
+++ b/store/multiversion/mvkv.go
@@ -316,7 +316,28 @@ func (v *VersionIndexedStore) VersionExists(version int64) bool {
 }
 
 func (v *VersionIndexedStore) DeleteAll(start, end []byte) error {
-	return v.parent.DeleteAll(start, end)
+	for _, k := range v.GetAllKeyStrsInRange(start, end) {
+		v.Delete([]byte(k))
+	}
+	return nil
+}
+
+func (v *VersionIndexedStore) GetAllKeyStrsInRange(start, end []byte) (res []string) {
+	keyStrs := map[string]struct{}{}
+	for _, pk := range v.parent.GetAllKeyStrsInRange(start, end) {
+		keyStrs[pk] = struct{}{}
+	}
+	for k, val := range v.writeset {
+		if val == nil {
+			delete(keyStrs, string(k))
+		} else {
+			keyStrs[string(k)] = struct{}{}
+		}
+	}
+	for k := range keyStrs {
+		res = append(res, k)
+	}
+	return res
 }
 
 // GetStoreType implements types.KVStore.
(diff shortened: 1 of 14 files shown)


store/multiversion/mvkv.go
```diff
@@ -316,7 +316,28 @@ func (v *VersionIndexedStore) VersionExists(version int64) bool {
...
+		v.Delete([]byte(k))
+	}
+	return nil
+}
+
+func (v *VersionIndexedStore) GetAllKeyStrsInRange(start, end []byte) (res []string) {
+	keyStrs := map[string]struct{}{}
+	for _, pk := range v.parent.GetAllKeyStrsInRange(start, end) {
+		keyStrs[pk] = struct{}{}
+	}
```
udpatil, line 329: we need to also consider the multiVersionStore when getting all the keys in range, since earlier txs potentially may have written within the range or deleted keys within the range, both of which should affect the result of this function. As such, we would also need to update readsets here, right? we should add all of these keys read both from parent OR from multiversion store to the readset so that the DeleteAll is properly backed by an updated readset representing the accesses of those keys / values

udpatil: actually, we should do something like iterateset in this case, because the keys that are accessed within the range is important here

udpatil: I think the easiest solution in this case (perhaps not the most optimal but should be correct) would be to create an Iterator within deleteAll, iterate over the whole iterator collecting keys, and THEN call delete on each one. Then you get the iteration safety out of the box

udpatil (changes requested):
updated comments from the earlier review clarifying further what needs to change in DeleteAll for OCC correctness

store/multiversion/mvkv_test.go
```diff
@@ -474,3 +474,28 @@ func TestRemoveLastEntry(t *testing.T) {
...
+		"key2": []byte("value8"),
+		"key3": nil,
+	})
+	mvs.SetWriteset(2, 1, map[string][]byte{
+		"key4": nil,
+	})
+	vis.Set([]byte("key5"), []byte("value7"))
+	vis.Delete([]byte("key6"))
+	require.Equal(t, []string{"key1", "key2", "key4", "key5"}, vis.GetAllKeyStrsInRange(nil, nil))
+}
```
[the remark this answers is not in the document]
codchen: added

PR #465, opened 2024-03-19
Rework abort handling to not be dependent on extracting codes from delivertx errors

udpatil:
This should appropriately allow for cases where panics are recovered and swallowed earlier because we now simply use the panic to abort execution, and detect that an abort occurred by reading the abort channel as the source of truth

Scheduler tests, and need to test alongside atlantic-2


stevenlanders (approved):
If not panicking, as long as there can only be a limited number of aborts (e.g., size of stores list), this is fine, otherwise the deliverTx may hang and we may have to go back to the `async deliverTx while draining abort channel` shape.

PR #468, opened 2024-03-20
[OCC] if no txs, avoid scheduler overhead, limit tasks

stevenlanders:
- avoids scheduler if no txs (mainly for gassless situations)
- limits tasks to the min(task count, workers)

- unit tests and load testing

### DIFF
--- a/baseapp/abci.go
+++ b/baseapp/abci.go
@@ -252,18 +252,22 @@ func (app *BaseApp) CheckTx(ctx context.Context, req *abci.RequestCheckTx) (*abc
 // DeliverTxBatch executes multiple txs
 func (app *BaseApp) DeliverTxBatch(ctx sdk.Context, req sdk.DeliverTxBatchRequest) (res sdk.DeliverTxBatchResponse) {
 	responses := make([]*sdk.DeliverTxResult, 0, len(req.TxEntries))
-	// avoid overhead for empty batches
-	if len(req.TxEntries) > 0 {
-		scheduler := tasks.NewScheduler(app.concurrencyWorkers, app.TracingInfo, app.DeliverTx)
-		txRes, err := scheduler.ProcessAll(ctx, req.TxEntries)
-		if err != nil {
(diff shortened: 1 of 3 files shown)


baseapp/abci.go
```diff
@@ -252,18 +251,18 @@ func (app *BaseApp) CheckTx(ctx context.Context, req *abci.RequestCheckTx) (*abc
...
-	scheduler := tasks.NewScheduler(app.concurrencyWorkers, app.TracingInfo, app.DeliverTx)
-	// This will basically no-op the actual prefill if the metadata for the txs is empty
-
-	// process all txs, this will also initializes the MVS if prefill estimates was disabled
-	txRes, err := scheduler.ProcessAll(ctx, req.TxEntries)
-	if err != nil {
-		// TODO: handle error
-	}
-
 	responses := make([]*sdk.DeliverTxResult, 0, len(req.TxEntries))
```
udpatil, line 254: is it safe to create a slice with capacity 0? i think its probably fine but just checking

baseapp/abci.go
```diff
@@ -252,18 +251,18 @@ func (app *BaseApp) CheckTx(ctx context.Context, req *abci.RequestCheckTx) (*abc
...
-	for _, tx := range txRes {
-		responses = append(responses, &sdk.DeliverTxResult{Response: tx})
+	// avoid overhead for empty batches
+	if len(req.TxEntries) > 0 {
```
udpatil, line 256: nit: slight preference for defensive checks with early return as opposed to nested ifs

udpatil (approved):

baseapp/abci.go
stevenlanders: good point - that's idiomatic golang

baseapp/abci.go
stevenlanders: yes it's safe (just didn't want it to be nil for marshaling reasons, which would be idiomatic golang)

PR #471, opened 2024-03-22
[staking] feat: add unbonding_id to UnbodingDelegationEntry

kyriediculous:

Add unbonding_id to protobuf message and regenerate the go bindings.

This is needed for https://github.com/sei-protocol/sei-chain/pull/1356


kyriediculous:
We probably also still want to change `MsgDelegateResponse` and `MsgUndelegateResponse` instead, which will make the precompile handlers much more efficient.

PR #479, opened 2024-03-29
[OCC] if synchronous, reset non-pending  (#477)

stevenlanders:
- Some tasks could be in waiting states, and this ensure those have been reset/incremented

- Unit tests


stevenlanders:
already included in different PR

PR #487, opened 2024-04-17
Cosmos gas multiplier params

Kbhat1:
- Add Cosmos gas multiplier numerator and denominator as param instead of hardcoded
- Corresponding sei-chain PR: https://github.com/sei-protocol/sei-chain/pull/1546
- NOTE: Adding more basic unit tests

- Sei-chain + verified on local chain

### DIFF
--- a/x/params/module.go
+++ b/x/params/module.go
@@ -100,6 +100,7 @@ func (am AppModule) RegisterInvariants(_ sdk.InvariantRegistry) {}
 func (am AppModule) InitGenesis(ctx sdk.Context, _ codec.JSONCodec, _ json.RawMessage) []abci.ValidatorUpdate {
 	defaultGenesis := types.DefaultGenesis()
 	am.keeper.SetFeesParams(ctx, defaultGenesis.FeesParams)
+	am.keeper.SetCosmosGasParams(ctx, defaultGenesis.CosmosGasParams)
 	return []abci.ValidatorUpdate{}
 }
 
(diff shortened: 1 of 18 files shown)


x/params/module.go
```diff
@@ -100,6 +100,7 @@ func (am AppModule) RegisterInvariants(_ sdk.InvariantRegistry) {}
 func (am AppModule) InitGenesis(ctx sdk.Context, _ codec.JSONCodec, _ json.RawMessage) []abci.ValidatorUpdate {
 	defaultGenesis := types.DefaultGenesis()
 	am.keeper.SetFeesParams(ctx, defaultGenesis.FeesParams)
+	am.keeper.SetCosmosGasParams(ctx, defaultGenesis.CosmosGasParams)
```
codchen, line 103: will this be called during an upgrade (instead of a re-genesis)?

Kbhat1: @codchen add in a migration

x/params/types/params.go
```diff
@@ -41,3 +44,30 @@ func validateFeesParams(i interface{}) error {
...
+func NewCosmosGasParams(multiplierNumerator uint64, multiplierDenominator uint64) CosmosGasParams {
+	return CosmosGasParams{
+		CosmosGasMultiplierNumerator:   multiplierNumerator,
+		CosmosGasMultiplierDenominator: multiplierDenominator,
+	}
+}
+
+func (cg *CosmosGasParams) Validate() error {
+	if cg.CosmosGasMultiplierDenominator == 0 {
+		return errors.New("Cosmos Gas Multiplier Denominator can not be 0")
```
udpatil, line 57: shouldnt we also assert that numerator can't be 0 to safeguard against making all txs free?

udpatil (approved):

x/params/types/params.go
Kbhat1: I intentionally left that out. Denominator not 0 is required to avoid panics. Numerator can validly be 0 and it requires a gov proposal to change. I think we can leave this as is.

Kbhat1: Update - added in the validation check here to avoid numerator is 0 after discussion

PR #494, opened 2024-04-26
Prevent abort channel from filling up

udpatil:
This prevents writing more than once to the abort channel for an instance of mvkv, which should prevent the abort channel from filling up.

### DIFF
--- a/store/multiversion/mvkv.go
+++ b/store/multiversion/mvkv.go
@@ -2,6 +2,7 @@ package multiversion
 
 import (
 	"bytes"
+	"fmt"
 	"io"
 	"sort"
 
@@ -92,7 +93,6 @@ type VersionIndexedStore struct {
 	incarnation      int
 	// have abort channel here for aborting transactions
 	abortChannel chan scheduler.Abort
-	abortWritten bool
 }
 
 var _ types.KVStore = (*VersionIndexedStore)(nil)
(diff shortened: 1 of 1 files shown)


store/multiversion/mvkv.go
```diff
@@ -122,6 +123,14 @@ func (store *VersionIndexedStore) GetWriteset() map[string][]byte {
 	return store.writeset
 }
 
+// WriteAbort writes an abort to the store but only allows one abort to be written PER instance of mvkv. This is because we pair abort channel writes with panics, and if we hit this more than once, it means that the panic was swallowed, so we won't write any aborts after a first abort is written to prevent any potential for deadlocking due to full channels
+func (store *VersionIndexedStore) WriteAbort(abort scheduler.Abort) {
+	if !store.abortWritten {
```
stevenlanders, line 128: unless there's a lock situation, i see this racing

stevenlanders:
```
select {
    case ch <- 2: // Put 2 in the channel unless it is full
    default:
        fmt.Println("Channel full. Discarding value")
    }
```
could we try this kind of shape?  This would prevent a 'full' situation at least

udpatil: updated to gracefully not write to channel

stevenlanders (approved):

PR #505, opened 2024-05-10
Fix export genesis for historical height

yzang2019:
Add a new function for baseapp LoadVersionWithoutInit so that we allow genesis export to export historical height after app.New is called.


storev2/commitment/store.go
```diff
@@ -94,6 +94,7 @@ func (st *Store) Set(key, value []byte) {
 
 // Implements types.KVStore.
 func (st *Store) Get(key []byte) []byte {
+	fmt.Printf("Getting key %s, current version is %d\n", string(key), st.tree.Version())
```
udpatil, line 97: we probably want to remove prints, right?

storev2/rootmulti/store.go
```diff
@@ -364,14 +366,9 @@ func (rs *Store) LoadVersionAndUpgrade(version int64, upgrades *types.StoreUpgra
...
 	}
-	if err := rs.scStore.Initialize(initialStores); err != nil {
-		return err
-	}
-	if version > 0 {
-		_, err := rs.scStore.LoadVersion(version, false)
-		if err != nil {
-			return nil
-		}
+	rs.scStore.Initialize(initialStores)
```
udpatil, line 369: this can still error, right? if so, don't we still need to return?

udpatil (approved):

storev2/rootmulti/store.go
yzang2019: This will not error after the seidb change

yzang2019: https://github.com/sei-protocol/sei-db/pull/64/files

PR #511, opened 2024-06-01
Add error for evm revert

udpatil:
This adds an error code for evm revert, and updates the tx result with it in the case of evm response containing a VM Error

Tested on local chain

### DIFF
--- a/types/errors/errors.go
+++ b/types/errors/errors.go
@@ -156,6 +156,9 @@ var (
 	// ErrUnsupportedTxType defines an error for an unsupported transaction type (e.g., EIP-4844 transactions)
 	ErrUnsupportedTxType = Register(RootCodespace, 44, "unsupported transaction type")
 
+	// ErrEVMVMError defines an error for an evm vm error (eg. revert)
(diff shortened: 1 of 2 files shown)


types/errors/errors.go
```diff
@@ -156,6 +156,9 @@ var (
 	// ErrUnsupportedTxType defines an error for an unsupported transaction type (e.g., EIP-4844 transactions)
 	ErrUnsupportedTxType = Register(RootCodespace, 44, "unsupported transaction type")
 
+	// ErrEVMVMError defines an error for an evm vm error (eg. revert)
+	ErrEVMVMError = Register(RootCodespace, 45, "evm vm error")
```
stevenlanders, line 160: What do you think about `evm reverted`?  (Often it's some contract validation that reverts, so it's not really a broken thing)

stevenlanders (approved):

codchen (approved):

types/errors/errors.go
udpatil: yeah sounds good

PR #519, opened 2024-06-21
Genesis export stream

jewei1997:
Writes genesis export to a file but with each module being appended to the file individually. We need this because we are running in an issue where the app state is too large to hold in memory causing an OOM. We currently assume that each individual module's app state is small enough to hold in memory.

manual testing locally.


server/export.go
```diff
@@ -38,6 +51,20 @@ func ExportCmd(appExporter types.AppExporter, defaultNodeHome string) *cobra.Com
...
+				return err
+			}
+
+			streamingFile, err := cmd.Flags().GetString(FlagStreamingFile)
+			if err != nil {
+				return err
+			}
+
+			if isStreaming && streamingFile == "" {
+				return fmt.Errorf("file to export stream to not provided")
```
yzang2019, line 65: Maybe add some instructions like please specify --streaming-file

server/start.go
```diff
@@ -386,13 +388,27 @@ func startInProcess(
 		config.GRPC.Enable = true
 	} else {
 		ctx.Logger.Info("starting node with ABCI Tendermint in-process")
+		var gen *tmtypes.GenesisDoc
+		if config.Genesis.StreamImport {
+			lines := genesistypes.IngestGenesisFileLineByLine(config.Genesis.ImportFile)
+			for line := range lines {
+				genDoc, err := tmtypes.GenesisDocFromJSON([]byte(line))
+				if err != nil {
+					continue
```
yzang2019, line 397: If there's error, we should just break out right? There's not much value to continue

PR #525, opened 2024-07-17
Add public interface for multistore to return earlist version

yzang2019:
This PR add a public interface for Multistore to return the earlist available version for state store (historical state). For seidb, if only SC is enabled, then it will assume only latest state is available.


Kbhat1 (approved):
LGTM, one question is if we can include the sc snapshot history in earliest version (like last 10k blocks) but I guess its not straightforward to find that history

PR #534, opened 2024-08-14
Stop executing the handler when proposal is sumbmitted

wojciech-sei:

CI will test it


x/gov/keeper/proposal.go
```diff
@@ -20,15 +20,6 @@ func (keeper Keeper) SubmitProposalWithExpedite(ctx sdk.Context, content types.C
 		return types.Proposal{}, sdkerrors.Wrap(types.ErrNoProposalHandlerExists, content.ProposalRoute())
 	}
 
-	// Execute the proposal content in a new context branch (with branched store)
-	// to validate the actual parameter changes before the proposal proceeds
-	// through the governance process. State is not persisted.
-	cacheCtx, _ := ctx.CacheContext()
-	handler := keeper.router.GetRoute(content.ProposalRoute())
-	if err := handler(cacheCtx, content); err != nil {
-		return types.Proposal{}, sdkerrors.Wrap(types.ErrInvalidProposalContent, err.Error())
```
philipsu522, line 29: does this have the side effect of allowing users to submit invalid props?

wojciech-sei: yes

cordt-sei:
Oops, button-mashed and updated this branch.

x/gov/keeper/proposal.go
yzang2019: Is it possible to fix the underline issue other than removing this handler? What does the upstream tendermint do here?

wojciech-sei: We could mark `ctx` with some flag saying that handler is executed in the "submit" scope and then act accordingly in the handler. But it means that we have to be careful about every handler in the system.

The current cosmos sdk stopped executing the handlers: https://github.com/cosmos/cosmos-sdk/blob/release/v0.50.x/x/gov/keeper/proposal.go#L72

They do it only for legacy types for backward compatibility, but no one uses those legacy proposals anymore.

yzang2019 (approved):

PR #538, opened 2024-09-04
Token allow list feature. Bank module changes

dssei:
This change is part of token allow list feature implementation for token extensions.
Added functions to:
- persist denom allow list and query it
- check if `from` and `to` addresses are in allowlist on bank send and multi-send

- unit testing
- functional testing (plugged modified code into sei-chain) and via ERC-20 pointer contract


x/bank/keeper/send.go
```diff
@@ -141,9 +155,45 @@ func (k BaseSendKeeper) InputOutputCoins(ctx sdk.Context, inputs []types.Input,
...
+		if !allowedToSend {
+			return sdkerrors.Wrapf(sdkerrors.ErrUnauthorized, "%s is not allowed to send funds", inAddress)
+		}
+	}
+	for _, out := range outputs {
+		outAddress, err := sdk.AccAddressFromBech32(out.Address)
+		if err != nil {
+			return err
+		}
+		allowedToReceive := k.IsAllowedToSendCoins(ctx, outAddress, out.Coins, denomToAllowListCache)
```
codchen, line 175: is it intended to use the same `IsAllowedToSendCoins` for receiver validation?

codchen (approved):
has one nit but overall lgtm

x/bank/keeper/send.go
dssei: Yeah, I was going on naming back and forth. But the same method is used for both sender and receiver validation.

PR #543, opened 2024-10-04
Refactor DeliverTx hook so that panics can be handled

codchen:
Panics in tx handler is usually handled by a defer statement containing a `recover()` clause. Previously the call stack looks like:
```
func DeliverTx:
    func runTx:
        deferred recover
        (actual processing)
    DeliverTx hook
```
In the above structure, DeliverTx hooks are run on `DeliverTx` level, so they are outside the deferred recover clause which is within `runTx` level.

This PR changes it to be:
```
func DeliverTx:
    func runTx:
        deferred recover
        (actual processing)
        DeliverTx hook
```
so that the hook can be recovered as well

unit test

### DIFF
--- a/baseapp/baseapp.go
+++ b/baseapp/baseapp.go
@@ -76,7 +76,7 @@ type (
 	// (or removed a substore) between two versions of the software.
 	StoreLoader func(ms sdk.CommitMultiStore) error
 
-	DeliverTxHook func(sdk.Context, sdk.Tx, [32]byte, abci.ResponseDeliverTx)
+	DeliverTxHook func(sdk.Context, sdk.Tx, [32]byte, sdk.DeliverTxHookInput)
 )
 
 // BaseApp reflects the ABCI application implementation.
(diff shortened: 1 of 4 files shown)


baseapp/baseapp.go
```diff
@@ -1007,6 +1007,11 @@ func (app *BaseApp) runTx(ctx sdk.Context, mode runTxMode, tx sdk.Tx, checksum [
 	if ctx.CheckTxCallback() != nil {
 		ctx.CheckTxCallback()(ctx, err)
 	}
+	resultStr := ""
+	res := app.getDeliverTxResponse(ctx, gInfo, result, &resultStr)
+	for _, hook := range app.deliverTxHooks {
+		hook(ctx, tx, checksum, res)
```
udpatil, line 1013: this will affect the actual gas Used for the tx execution, but will be using the `res` with a stale gas used from getDeliverTxResponse. is there a scenario where this difference between the deliverTx response used for hooks and deliver tx response returned via tendermint causes an issue?

codchen: hooks (as of right now) don't use gasUsed passed in. I changed the data structure to be one specific to delivertx hooks that exclude gasUsed to avoid future misuse.

baseapp/baseapp.go
```diff
@@ -1007,6 +1007,25 @@ func (app *BaseApp) runTx(ctx sdk.Context, mode runTxMode, tx sdk.Tx, checksum [
...
+	if ctx.IsEVM() {
+		evmTxInfo = &abci.EvmTxInfo{
+			SenderAddress: ctx.EVMSenderAddress(),
+			Nonce:         ctx.EVMNonce(),
+			TxHash:        ctx.EVMTxHash(),
+			VmError:       result.EvmError,
+		}
+	}
+	var events []abci.Event = []abci.Event{}
+	if result != nil {
```
udpatil, line 1020: don't we also need to handle the case in which `result.EvmError` is non-nil? or do we want to index events for that case as well?

codchen: the existing behavior is to index events even if EvmError is non-nil https://github.com/sei-protocol/sei-cosmos/pull/543/files#diff-389d02f089ab55575400f5300d3c5cfe277fde5f04bd0fe6a85456a7252d1099R309-R330

not sure if we want to change that. I can see reasons both ways

udpatil (approved):

PR #550, opened 2024-11-01
Move allow list logic to msg server

dssei:
Move allow list logic to msg server

- Unit tests
- Local chain functional tests

### DIFF
--- a/x/bank/keeper/send.go
+++ b/x/bank/keeper/send.go
@@ -59,26 +59,22 @@ type BaseSendKeeper struct {
 	paramSpace paramtypes.Subspace
 
 	// list of addresses that are restricted from receiving transactions
-	blockedAddrs map[string]bool
-	// map of all registered module addresses
-	allModuleAddresses map[string]bool
-	recipientCheckers  *[]RecipientChecker
+	blockedAddrs      map[string]bool
+	recipientCheckers *[]RecipientChecker
 }
 
 func NewBaseSendKeeper(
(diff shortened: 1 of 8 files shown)


x/bank/keeper/send.go
```diff
@@ -533,6 +534,19 @@ func (k BaseSendKeeper) IsAllowedToSendCoins(ctx sdk.Context, addr sdk.AccAddres
 	return true
 }
 
+func (k BaseSendKeeper) getAllModuleAddresses(ctx sdk.Context) []string {
```
dssei, line 537: Not the most efficient way to retrieve all module addresses and we may also want to cache it. So, I will look over the weekend how to improve this. I have also found we are passing some module addresses via `blockedAddrs map[string]bool`, but not all of them. And there is also module.Manager that may return them. @philipsu522 @codchen let me know what would the most efficient way.

philipsu522: yea i agree this might be too expensive. What about iterating through all modules (maybe app.go could expose this because we register all modules) and then calling https://github.com/sei-protocol/sei-cosmos/blob/63ac1b222e281be5ff25623ade4b93bd61517c1f/x/auth/keeper/keeper.go#L160 on it? Not sure if @codchen has a better idea here though

dssei: Refactored, to avoid module checks altogether.
I think for this feature we should only be in business of controlling transfers for regular account/addresses, not modules, with transfers to modules implicitly not allowed via `send`. And we should not interfere with module->module, account->module and module->account flows. 
The way to do this would be to move the allowlist logic to the message server and only do the checks in `send` and `multisend`.

codchen (approved):

philipsu522 (approved):

PR #558, opened 2024-12-27
Integrate with MinTxsInBlock

codchen:
Pass through MinTxsInBlock between app side and tendermint side

integrated with sei-chain and tested on local sei

### DIFF
--- a/baseapp/params.go
+++ b/baseapp/params.go
@@ -48,6 +48,10 @@ func ValidateBlockParams(i interface{}) error {
 		return fmt.Errorf("block maximum gas must be greater than or equal to -1: %d", v.MaxGas)
 	}
 
+	if v.MinTxsInBlock < 0 {
+		return fmt.Errorf("block min txs in block must be positive: %d", v.MinTxsInBlock)
+	}
+
 	return nil
 }
 
--- a/baseapp/params.go
+++ b/baseapp/params.go
@@ -49,7 +49,7 @@ func ValidateBlockParams(i interface{}) error {
 	}
 
 	if v.MinTxsInBlock < 0 {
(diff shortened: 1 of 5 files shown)


baseapp/params.go
```diff
@@ -48,6 +48,10 @@ func ValidateBlockParams(i interface{}) error {
 		return fmt.Errorf("block maximum gas must be greater than or equal to -1: %d", v.MaxGas)
 	}
 
+	if v.MinTxsInBlock < 0 {
+		return fmt.Errorf("block min txs in block must be positive: %d", v.MinTxsInBlock)
```
philipsu522, line 52: nit - "non-negative"

server/export.go
```diff
@@ -116,8 +116,9 @@ func ExportCmd(appExporter types.AppExporter, defaultNodeHome string) *cobra.Com
 					InitialHeight: exported.Height,
 					ConsensusParams: &tmtypes.ConsensusParams{
 						Block: tmtypes.BlockParams{
-							MaxBytes: exported.ConsensusParams.Block.MaxBytes,
-							MaxGas:   exported.ConsensusParams.Block.MaxGas,
+							MaxBytes:      exported.ConsensusParams.Block.MaxBytes,
+							MaxGas:        exported.ConsensusParams.Block.MaxGas,
+							MinTxsInBlock: exported.ConsensusParams.Block.MinTxsInBlock,
```
philipsu522, line 121: do we need a migration handler for thsi?

baseapp/params.go
codchen: updated

server/export.go
codchen: no, just a param update proposal

PR #564, opened 2025-02-19
Add whitelist for fee denoms

codchen:
It is possible for tx sender to include "garbage" tokens as fees which may decrease chain performance if in large quantity (in terms of the number of different denoms, which may be created via tokenfactory). This PR introduces a new parameter to whitelist denoms that can be used for fees. Note that `usei` is always whitelisted.

local sei with tokenfactory tokens


x/auth/ante/validator_tx_fee.go
```diff
@@ -19,13 +19,14 @@ func CheckTxFeeWithValidatorMinGasPrices(ctx sdk.Context, tx sdk.Tx, simulate bo
 	}
 
 	feeCoins := feeTx.GetFee()
+	feeParams := paramsKeeper.GetFeesParams(ctx)
+	feeCoins = feeCoins.NonZeroAmountsOf(append([]string{sdk.DefaultBondDenom}, feeParams.GetAllowedFeeDenoms()...))
```
philipsu522, line 23: do we need a migration handler for the new feeParams?

codchen: we don't need migration, since fees are marshal/unmarshaled via json so adding a new field won't break the compatibility. Added a test to verify this

philipsu522 (approved):

PR #593, opened 2025-07-31
[SEI-9449] Remove unused invariant checks

jewei1997:


x/distribution/keeper/invariants.go
```diff
@@ -1,162 +0,0 @@
...
-		}
-		res, stop = NonNegativeOutstandingInvariant(k)(ctx)
-		if stop {
-			return res, stop
-		}
-		res, stop = ReferenceCountInvariant(k)(ctx)
-		if stop {
-			return res, stop
-		}
-		return ModuleAccountInvariant(k)(ctx)
```
philipsu522, line 38: do we somehow capture these scenarios via unit tests?

PR #598, opened 2025-08-20
Integrate `GetTxPriority` ABCI interface

masih:
Add hooks to allow injection of transaction prioritization logic and handle ABCI requests to get transaction priority via the hook.

Relates to:
 * https://github.com/sei-protocol/sei-tendermint/pull/301

### DIFF
--- a/tasks/scheduler.go
+++ b/tasks/scheduler.go
@@ -256,18 +256,6 @@ func allValidated(tasks []*deliverTxTask) bool {
 	return true
 }
 
-func (s *scheduler) PrefillEstimates(reqs []*sdk.DeliverTxEntry) {
-	// iterate over TXs, update estimated writesets where applicable
-	for _, req := range reqs {
-		mappedWritesets := req.EstimatedWritesets
-		// order shouldnt matter for storeKeys because each storeKey partitioned MVS is independent
-		for storeKey, writeset := range mappedWritesets {
-			// we use `-1` to indicate a prefill incarnation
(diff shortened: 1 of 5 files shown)


baseapp/abci.go
```diff
@@ -1228,3 +1228,50 @@ func (app *BaseApp) LoadLatest(ctx context.Context, req *abci.RequestLoadLatest)
...
+	if tx == nil {
+		return nil, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "nil tx")
+	}
+	if err := tx.ValidateBasic(); err != nil {
+		return nil, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, err.Error())
+	}
+
+	// TODO: should we bother validating the messages here?
+	msgs := tx.GetMsgs()
+	if err := validateBasicTxMsgs(msgs); err != nil {
```
yzang2019, line 1263: This might be consuming some resources, but we can test it out to see if that make a difference or not

types/context.go
```diff
@@ -241,12 +242,19 @@ func (c Context) StoreTracer() gaskv.IStoreTracer {
 	return c.storeTracer
 }
 
-// WithEventManager returns a Context with an updated tx priority
+// WithPriority returns a Context with an updated tx priority.
 func (c Context) WithPriority(p int64) Context {
```
yzang2019, line 246: Where is this function being called?

baseapp/abci.go
masih: Right. I was on the fence for adding this too 👍

PR #606, opened 2025-09-11
refactor: use strings.builder

mickychang9:

strings.Builder has fewer memory allocations and better performance.
More info: [golang/go#75190](https://github.com/golang/go/issues/75190)


mickychang9:
@philipsu522 @sei-will @stevenlanders Hi, Could you please review this PR at your convenience? Thank you very much.

